PR #56, opened 2025-01-10
8 funkydate fix

### LINKED ISSUE #8: Bug: Funky Date
Angela Chen and Ida Voong

Description:
- Look at the info tab and look at the date
- BUG: day and month can be zero, which is confusing 
- FIX: make the day and month start at 1 (i.e 1/1/0 0 o'clock)
- NOTE: format is day/month/year

idavoong:
Group Members: Angela Chen, Ida Voong
### What changed:
- Changed `packages\server\src\date\fantasyDate.ts` so that when a new fantasy date object is created, the month and day start at 1 instead of 0.
- Added a new unit test to test this functionality in `packages\server\test\date\fantasydate.test.ts`
- Edited `packages/server/test/testSetup.ts` so that we can run tests in parallel by using different database instances. Running tests on the same database can cause flaky tests since it's trying to edit data at the same time.
- [Closes #8 ](https://github.com/sloalchemist/potions/issues/8)

### Why?:
- It's confusing to have day and month start at 0.


anvch:
LGTM

anvch (commented):
LGTM

packages/server/test/testSetup.ts
```diff
@@ -10,11 +10,11 @@ export let world: ServerWorld;
 export let village: Community;
 export let itemGenerator: ItemGenerator;
 
-export const commonSetup = () => {
+export const commonSetup = (testName: string) => {
   // Any common setup code
   jest.clearAllMocks();
 
-  initializeServerDatabase('data/test.db', true);
+  initializeServerDatabase(testName, true);
```
StarDylan, line 17: Could you add a bit of context in the PR about why this change was made?

Also, just a friendly reminder to use the [Conventional Commit](https://www.conventionalcommits.org/) format as outlined in the `CONTRIBUTING.md`.

StarDylan (changes requested):

packages/server/test/testSetup.ts
anvch: This change was made after discussing with the professor because we use pnpm to test. Before, our tests were failing because it was trying to access the same resource (data/test.db) at the same time. The testName parameter adds a field so that when you setup the test, you can create a separate instance of the server database for each specific test file.

By adding context in the code, do you just mean adding a comment explaining the format of testName and its purpose? Also, the conventional commit format is noted - is it ok to just use that format moving forward because we can't change our previous commit messages?

StarDylan: > By adding context in the code, do you just mean adding a comment explaining the format of testName and its purpose?

Yes. But I recommend using a JSDoc for `commonSetup` like this:
```
/**
 * Add 2 numbers
 * @param a The first number
 * @param b The second number
 */
  return a + b;
}
```
Then we get nice documentation when hovering over it:

image

Then in a separate line comment, you can describe why we need this parameter at all (to avoid concurrent access issues since jest runs test files in parallel).

Here's a nice paragraph from Google describing the difference between line comments and JSdocs https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments

> Also, the conventional commit format is noted - is it ok to just use that format moving forward because we can't change our previous commit messages?

Yes that's fine.

StarDylan (approved):
Awesome, looks great! Thank you for implementing this for the class. 

For future reference, using the prefix `docs:` might be a slightly better fit for the commit message, but the format looks good.

PR #58, opened 2025-01-10
Make cauldron not walkable

jmdebarro:
Group: Logan Costello, Jonathan DeBarro

Changed configuration files to make cauldron not walkable.
`packages/client/static/global.json`
`packages/server/static/global.json`

Added unit test to see if item is walkable.
`packages/server/test/items/itemsWalkable.test.ts`

Edited a testSetup.ts file to take in an optional string that becomes the database name for the testing instance. If nothing is passed, 'data/test.db' is assigned.
`packages/server/test/testSetup.ts`


jackalnom:
To confirm, you tested to make sure you can still brew potions at a cauldron after this change?

jmdebarro:
> To confirm, you tested to make sure you can still brew potions at a cauldron after this change?

Actually no we have to write that unit test. We just checked that we could do it through the game, which we could

StarDylan:
Next time make sure to use the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. But otherwise it looks good! I'm waiting on merging until #56 is merged, since they're working on some documentation for `testSetup.ts`

StarDylan:
@jmdebarro and @logancostello Please fix the merge conflict

jmdebarro:
> @jmdebarro and @logancostello Please fix the merge conflict

We need to write another test case. I'll ping you @StarDylan when it's fully complete. Apologies for opening this PR early.

logancostello:
> To confirm, you tested to make sure you can still brew potions at a cauldron after this change?

Would it be better create a test that you can interact with an item while next to it (rather than on top of it) if a) the item is not walkable and b) the item has an interaction. This would make it more generic, a bit easier to test, and also help test potion shops (we made the same change to them as well).

logancostello (commented):
LGTM. In the future we should use unit tests rather than manual test to test that the cauldron and potion stand can be still be interacted with.

jmdebarro:
Hey @StarDylan We're good now. Prof said we didn't need that second unit test, so this should be fine. Let us know if there is anything we should change. Thanks for hanging with us.

logancostello (commented):
Still looks good from the last time

logancostello (approved):

StarDylan (approved):
Great, it looks good!

PR #59, opened 2025-01-11
39 BUG: Negative Health Possible for Mobs

xgarcia17:
Group Members: Xavier Garcia, Kaila Aquino, Dre Seguitan
### The Bug
Before the fix, when mobs are attacked, they lose health. When mobs take on more damage points than they have health points remaining, their health becomes a negative number. Instead, in this instance, the mob's health should be set to zero.
### What Changed
- Changed `packages/server/src/mobs/mob.ts` in the function `changeHealth()`, in addition to checking the upper-bound of a mob's health with the `maxHealth` property, code was added to check the lower-bound of a mob's health so that it does not become less than zero.
- Created the directory `packages/server/test/mobs` along with the test file `mobNegativeHealth.test.ts` which is contained inside said directory. The test file tests the identified bug and it's fix.
- Edited `packages/server/test/testSetup.ts` so that we can run tests in parallel by using different database instances, and so that we can create the necessary entities required for spawning mobs.


StarDylan:
Next time make sure to use the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. But otherwise it looks good! I'm waiting on merging until #56 is merged, since they're working on some documentation for `testSetup.ts`

StarDylan:
Hey @kailaaquino and @xgarcia17 , could you please resolve the merge conflict using the instructions I shared on Slack before I review? Let me know if you have any questions

kailaaquino:
Hi @StarDylan I just merged and pushed the changes to our branch.

packages/server/test/mobs/mobNegativeHealth.test.ts
```diff
@@ -0,0 +1,151 @@
+import { commonSetup, graph } from '../testSetup';
+import { mobFactory } from '../../src/mobs/mobFactory';
+import { Mob } from '../../src/mobs/mob';
+import { DB } from '../../src/services/database';
+import { Community } from '../../src/community/community';
+import { buildAndSaveGraph, constructGraph, initialize } from '@rt-potion/converse';
+
+beforeAll(() => {
+  commonSetup("mobNegativeHealth");
```
StarDylan, line 9: Also please update this to use the new `commonSetup`'s parameter format: `data/[TEST_FILENAME].db`

kailaaquino: `commonSetup("data/mobNegativeHealth.test.db");`
Would it be like this?

StarDylan: Yep that works

kailaaquino: Just pushed the changes

packages/converse/data/test.db
StarDylan, line 1: These files are artifacts of the build and should be removed with:
```
git rm --cached packages/converse/data/test.db
git rm --cached packages/server/itemGenerator
git rm --cached packages/server/mobNegativeHealth
git rm --cached packages/server/mobNegativeHealth.test
```

StarDylan (changes requested):
Everything looks good, just need to remove these files and you'll be all set!

StarDylan (approved):
Looks great!

PR #61, opened 2025-01-11
BUG 27: Non-Allied Mobs Passing Items Between One Another

AndrewJMart:
Group Members: Andrew Martinez, Jasmine Cabrera, Allen Choi

The Bug

Nonallied mobs, such as the player and blob, are able to pass items to one another. Specifically, a player would be able to give items such as a log to a blob. This can be abstracted into nonallied mobs passing items between each other. 

What Changed

\potions\packages\server\src\items\carryable.ts  giveItem()
- Added inline comments to document the function
- Added if statement to check if mobs are allied before passing

\potions\packages\server\src\community\community.ts static isNotAllied()
- Added doc string describing Function
- Function checks alliance database to assert whether two mobs are in an alliance or not


TheAllenChoi:
lgtm

StarDylan:
Looks good! We just need to remove the test.db file with (then it should be ignored with the .gitignore):
```
git rm --cached packages/server/data/test.db
```

(also friendly remember to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format.

StarDylan (changes requested):
Looks good! We just need to remove the test.db file with (then it should be ignored with the .gitignore):
```
git rm --cached packages/server/data/test.db
```

Also friendly remember to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format.

AndrewJMart:
Test DB should be deleted!

StarDylan (approved):
Great, it looks good!

Issue #35, opened 2025-01-08
Bug: When player switches tabs after a game over, the screen freezes (Kaatje, Emily)


emlai1263:
![image](https://github.com/user-attachments/assets/09f1dc58-02fb-4c6e-9a93-3e875c86adf0)

PR #62, opened 2025-01-12
Bug: Game freezes on blur (blue screen)

emlai1263:
Group Members: Emily Lai, Kaatje Matthews

# Bug: Issue #35 
When the player gets a game over, and the unfocuses the game (switching tabs, clicking outside the window), the blur won't go away, regardless of focus status

# What changed
- the function handleDestoryMob in packages\client\src\services\serverToBroadcast.ts has a Typescript Promise that makes the game wait for the game is in focus before following through with disconnecting them to the world, so the screen freezes on the Game Over instead of the blur
- in main.ts, the game can't blur if the player has died
- packages\client\src\services\serverToBroadcast.ts - added playerDead boolean, which main.ts uses to prevent blurring if game over is reached
- packages\client\src\main.ts - added focused boolean, to let serverToBroadcast.ts see if the game is blurred or not

* Note on unit testing - This bug is more UI based and isn't good for unit testing, so no unit test is included (This has been discussed with Prof Pierce)


packages/client/src/scenes/worldScene.ts
```diff
@@ -437,4 +437,5 @@ export class WorldScene extends Phaser.Scene {
     text.setScrollFactor(0); // Make it stay static
     text.setDepth(100);
   }
+  // remove blur?
```
StarDylan, line 440: What's this comment for?

emlai1263: Woops, that was a note to myself when I was looking through the code, I removed it!

StarDylan (approved):
Looks good!

PR #63, opened 2025-01-12
Validate User Name Inputs

AdamMikacich:
**Group:** Adam Mikacich, Ananya Thapar, Caroline Koddenberg

**Bug description:**
Prior to this PR, usernames had no restrictions. Names longer than 10 characters would overflow off the screen.

**What changed:**
We added a new `validators.ts` file to hold functions used for parsing user input. While this new file only contains our parseName function for now, it will come in handy for additional parses (such as those for chat input) as development continues. After a user enters their newly desired name, the parseName function trims the input and ensures the result is between 1 and 10 characters inclusive. If the new name does not satisfy these requirements, then the user is notified of these requirements (in which they can try again afterward).

**Why:**
Users expect their username to fit nicely on screen. Therefore, we must limit their input to a reasonable size to fit within the game. Moreover, whitespaces are trimmed before calculating the length and saving the result to remove accidental spaces before/after the name.

**Before:**
Screenshot 2025-01-11 at 6 18 47 PM
**After:**
Screenshot 2025-01-11 at 6 19 28 PM


StarDylan (approved):
Looks great! Good job!

Issue #40, opened 2025-01-08
Bug: Cannot chat again with same NPC until you go out and back into range. (Darren Hong, Andrew Mei)


DarrenH6293:

Andrew76xx:
Works if refreshed.

PR #66, opened 2025-01-12
fix: rechatting with npcs

Andrew76xx:
**Group**: Darren Hong, Andrew Mei

**Bug Description** (Issue #40): When finishing a conversation (chatting) with an NPC, the player cannot talk with the same NPC unless the nearby Mobs List changes.

**What Changed**: We added a line of code in `controller.ts` for the `setChatting` function, where the list of `lastChatCompanions` is reset every time a conversation is finished. The function, `mobRangeListener`, compares the nearby mobs with the mobs the player has recently chatted with and resets the `lastChatCompanions` list whenever the conversation ends. This ensures that the same NPCs can be re-added to the chat companion list, even if they were recently interacted with, as long as they remain in range.

**Why**: Previously, to chat with the same NPC, the player would have to go out of the mob's range and get back in OR wait for a new mob to be near the player to update the list of nearby mobs. By solving this issue, the player does not have to take unnecessary actions to chat with the same NPC.

Before:
image

After:
image

### DIFF
--- a/packages/client/src/world/controller.ts
+++ b/packages/client/src/world/controller.ts
@@ -45,6 +45,10 @@ export function setGameState(state: GameState) {
 export function setChatting(chat: boolean) {
   console.log('setChatting', chat);
   chatting = chat;
+  // Allows for rechatting with the same NPC
+  if (!chat) {
+    lastChatCompanions = [];
+  }
 }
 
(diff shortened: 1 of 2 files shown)


StarDylan (approved):
Looks good!

PR #68, opened 2025-01-12
fix: CI failing on client tests

StarDylan:
## Bug

The CI was failing on main due to jest not being set up to mock the DOM / canvas.

## Changes:
[fix(tests): move mocks into beforeEach](https://github.com/sloalchemist/potions/commit/6a0748dc56e89be288e7cc2f2688307bb14f657e)

- There was an issue "TypeError: Cannot read properties of undefined (reading 'Container')" when setting up the mocks outside of the beforeEach.

[chore(client): switch jest to jsdom environment](https://github.com/sloalchemist/potions/commit/11b6a0058e417d25d7b55e7c0b149e5ebe7375c5)
- Some tests require mocking the canvas, which requires test to run in the jsdom environment. This was set as the default for all tests in client

### DIFF
--- a/packages/client/jest.config.mjs
+++ b/packages/client/jest.config.mjs
@@ -1,10 +1,11 @@
 export default {
   preset: 'ts-jest',
-  testEnvironment: 'node',
+  testEnvironment: 'jsdom',
   moduleFileExtensions: ['ts', 'js'],
   transform: {
     '^.+\\.ts$': 'ts-jest',
   },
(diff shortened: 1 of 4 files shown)


jackalnom (approved):
This is fine for now, I think this is a symptom of some of the client code not being properly abstracted though. Ideally, you'd break out anything unit testable so it doesn't rely on anything phaser or browser related.

Issue #23, opened 2025-01-06
Bug: You can add a blue potion to a red potion shop and it makes it a red potion (McCay, Charlotte, Shawn) 


Mcray3000:
When I ran out of red potions and added a blue potion to my stand it became red. This does the same in reverse, i.e. adding a red potion to a blue stand turns the red potion into a blue one.

jackalnom:
good one...

PR #69, opened 2025-01-12
Fix: Restrict Potion Stands to Accept Only Matching Potion Types

sgill33:
## Group: Shawn Gill, McCay Ruddick, Charlotte Maples

**Closes:** #23 

## Summary:
Fixed bug where after creating a blue potion stand, you could add a red potion to it and vice verse. With our changes, after create a certain stand subtype, it will only accept potions with the same subtype. 

Ex: Player creates a potion stand from a red potion, a potion stand is created with a red potion sprite. If this player brews another red potion, they can add it to the red potion stand. If the player tries adding a non-red potion to the stand, nothing will happen and they will still be carrying the potion.

## Changes:
- `packages/server/src/items/container.ts` :  Added if statement to ensure potion and potion-stand subtype match before adding a potion. This was the main bug fix
- `packages/server/test/items/addPotionToStand.test.ts` : Created tests to ensure blue potion goes into a blue potion stand, and a red potion does not go into a blue potion stand
- `packages/server/test/testSetup.ts` : Added a potion stand as a item type


Mcray3000 (approved):
Yay :)

packages/converse/data/test.db
StarDylan, line 1: This file is an artifact of the build and should be removed with:
```
git rm packages/converse/data/test.db
```

StarDylan (changes requested):
Also just a friendly remember to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format for your future changes.

charm-13 (approved):

StarDylan (approved):
Looks good!

PR #73, opened 2025-01-13
Fix: don't display alert when cancel is pressed

ruiyichi:
**Group:** Daniel Ching, Josh Queja

### Bug
On the start screen, clicking cancel when the character name window appears displays the invalid name alert.

### Changes
https://github.com/sloalchemist/potions/commit/8841c1482e11ba02abeedd5cb856facc675c9011

- If the window prompt is null, exit the function early and do not display alert

### Testing
- UI change only, no unit test needed


StarDylan (approved):
Looks great!

PR #75, opened 2025-01-13
Fix: Wall Can be Built From Partial Wall (Issue 13)

jhuanny:
Fix: Wall Can be Built From Partial Wall 

## Summary:
Fixed bug where player carrying log interacting with a partial wall would not create a full wall segment. With our changes, a complete wall is able to be created in place of the partial wall. 

## Changes:
- `packages/server/src/items/uses/building/buildWall.ts` : Changed if statement to check if partial wall 'complete' attribute is _greater_ than zero. This was the main bug fix.
- `packages/server/test/items/uses/building.test.ts` : Created tests to ensure wall can be successfully built from a partial wall by a mob carrying a log


halliec6 (approved):

packages/converse/data/building.test.db
StarDylan, line 1: These files are artifacts of the test and should be removed with:
```
git rm --cached packages/converse/data/building.test.db
git rm --cached packages/converse/data/itemGivingtest.db
git rm --cached packages/converse/data/mobNegativeHealthtest.db
```

rachelHartfelder (approved):

StarDylan (changes requested):
Overall looks good, just need to remove some extraneous files. Nice job with the Conventional Commit messages

StarDylan (approved):
Looks good!

PR #97, opened 2025-01-13
Remove all package-lock.json files as they conflict with pnpm

AdamMikacich:
**Group:** Adam Mikacich, Ananya Thapar, Caroline Koddenberg

**Bug description:**
All `package-lock.json` files should be removed as they are leftover from running `npm install`. Pnpm already manages its own lock file (`pnpm-lock.yaml`) in the root of the repository. This lock file is used for all workspace packages. Keeping `package-lock.json` files while simultaneously using pnpm for package management will lead to dependency discrepancies (ultimately causing "it works on my machine" issues). Fixing this will ensure there is one source of truth, pnpm.

**What changed:**
Removed all `package-lock.json` files and added `package-lock.json` to the `.gitignore` to prevent further native npm lock files from being added to the repo.


StarDylan (approved):
Awesome, this looks all good

PR #98, opened 2025-01-13
Fix: Issue 44 - mobs already pathing into the fenced area do not respect player fences (Ryan Hu, Ryan Chan, Kyle Taschek)

### LINKED ISSUE #44: Bug: mobs already pathing into the fenced area do not respect player fences (Ryan Hu, Ryan Chan, Kyle Taschek)

ryanchansf:
Closes #44 

### Summary
This PR resolves the bug where mobs with an already calculated path can pass through newly created walls. Our fix checks at every tick whether the next step in the mob's path is occupied by an unwalkable item in the world's grid. If it is, the mob recalculates its path to avoid the obstacle.

NOTE: branch name indicated issue 49 should actually by 44

### Changes
`/client/src/world/mob.ts`
- `tick()`: We add additional checks and logic for regenerating the mob's path in the world. Utilizing the mob's path attribute and the `isWalkable()` method of the item at the next step, we determine if the mob is able to proceed or if it needs to recalculate its path.

### Testing
`/client/test/world.mob.test.ts`
- Unit tests for the new logic in the `tick()` method.
  - Test the mob's constructor to ensure that it can be created in the world.
  - Test the mob's path recalculation when the next step is blocked by an unwalkable item, mocking the world's grid and the mob's path.
  - Test that the mob's path is not recalculated when the next step contains a walkable item.
  - Test that the mob's path is not recalculated when the next step does not contain any item.
- 100% statement and branch coverage on our changes

We also ran the modified pathing logic in our local environment. The test involved building a new wall in front of the village gates and observing the mob's behavior. The mobs approaching the gates successfully recalculated its path to avoid the newly created wall rather than phasing through it, which confirms the fix.

### DIFF
--- a/packages/client/src/world/mob.ts
+++ b/packages/client/src/world/mob.ts
@@ -30,6 +30,7 @@ export class Mob extends Physical {
     this.maxHealth = maxHealth;
 
     if (position) {
+      console.log('Adding mob to grid', this.key, this.position);
       world.addMobToGrid(this);
     }
 
@@ -56,7 +57,7 @@ export class Mob extends Physical {
   }
 
   changePosition(world: World, newPosition: Coord) {
-    //console.log('changePosition', this.position, newPosition);
+    // console.log('changePosition', this.position, newPosition);
     world.moveMob(this, newPosition);
     this.position = newPosition;
   }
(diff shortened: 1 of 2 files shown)


StarDylan (approved):
Looks good! Also just a friendly remember to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format for future commits.

ryanhu021:
Will do, thanks Dylan!

PR #101, opened 2025-01-14
Bug 28 potion stand dropping

ddharton:
Bug 28

Potion Stand not dropping items when smashed. 

Fix:
Added functionality for potion stand to drop items when smashed.

Testing:
Created a test to check if the items would properly drop.

@AkashPraveen @ivannghi @ddharton

### DIFF
--- a/packages/server/src/items/smashable.ts
+++ b/packages/server/src/items/smashable.ts
@@ -39,8 +39,6 @@ export class Smashable {
 
   destroyPotionStand(){
     const gold = this.item.getAttribute<number>('gold');
-    console.log(gold);
-    //const itemsCount = this.item.getAttribute<number>('items');
     if (this.item.getAttribute<number>('gold') > 0){
       const position = Item.findEmptyPosition(this.mob.position);
       itemGenerator.createItem({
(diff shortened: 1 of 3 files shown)


packages/server/src/items/smashable.ts
```diff
@@ -16,16 +17,48 @@ export class Smashable {
     return undefined;
   }
 
+  private mob!: Mob; // Add mob as a class property
```
StarDylan, line 20: I'm a bit confused why you need the smashing mob's location. Wouldn't the items drop where the potion stand was smashed?

StarDylan (changes requested):

packages/server/src/items/smashable.ts
ivannghi: We used it for the position since it should be the same as the potion stand, and it is something that we can fix in our next ticket when we make it more abstract to work with all containers. We intended to use the position of the potion stand but found it easier in the moment to do it using the mob's location.

packages/server/src/items/smashable.ts
```diff
@@ -16,16 +17,48 @@ export class Smashable {
...
+      }
       this.item.destroy();
     }
   }
-}
+
+  destroyPotionStand(){
+    const gold = this.item.getAttribute<number>('gold');
+    console.log(gold);
+    //const itemsCount = this.item.getAttribute<number>('items');
```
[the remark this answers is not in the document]
ddharton: Removed Debug Lines

StarDylan (approved):
Looks good! Looking forward to the next ticket where `Smashable` becomes more abstract

PR #107, opened 2025-01-14
Feat: Blue potions no longer decrease health, but rather increase player speed

eswenke:
## Summary
- Changed the functionality of the blue potion from simply removing health. Now increases player speed. Also works with any mob who purchases and consumes a blue potion. Caps speed at 10 (which is REALLY fast), and is currently permanent until rebuild of game to refresh state.

## Changes
- `packages/server/src/mobs/mob.ts` Added a 'changeSpeed' method.
- `packages/server/src/items/potionEffects.ts`  Added a call to 'changeSpeed' to replace the old 'changeHealth'. Adds 2 speed to the player, capping at 10.
- `packages/server/src/services/clientCommunication/pubsub.ts`  Added 'changeSpeed' signature.
- `packages/server/src/services/clientCommunication/stubbedPubSub.ts` Added 'changeSpeed' signature.
- `packages/server/src/services/clientCommunication/ablyService.ts` Added 'changeSpeed' method for broadcasting mobChange.
- `packages/server/src/services/clientCommunication/clientMarshalling.ts` Added speed as an attribute to the player rather than a regular property.
- `packages/client/src/world/mob.ts` Comment speed out of the properties and parameters, as it is now an attribute.
- `packages/client/src/sprite/sprite_mob.ts` Comment speed out of constructor
- `packages/client/src/scenes/uxScenes.ts` Added speed to the current ux attributes shown in game (doesn't update in realtime yet)
- `packages/client/src/worldMetadata.ts` Added speed getter.
- `packages/common/src/mob.ts` Commented out speed.

## Added
- `packages/server/test/items/potionEffects.test.ts` Added unit testing.

### DIFF
--- a/packages/client/src/sprite/sprite_mob.ts
+++ b/packages/client/src/sprite/sprite_mob.ts
@@ -47,7 +47,6 @@ export class SpriteMob extends Mob {
       mob.type,
       mob.maxHealth,
       mob.position,
-      // mob.speed,
       mob.attributes
     );
     this.scene = scene;
--- a/packages/client/src/sprite/sprite_mob.ts
+++ b/packages/client/src/sprite/sprite_mob.ts
@@ -1,3 +1,4 @@
+
 import { MobI } from '@rt-potion/common';
 import { WorldScene } from '../scenes/worldScene';
 import { SpriteItem } from './sprite_item';
@@ -46,7 +47,7 @@ export class SpriteMob extends Mob {
       mob.type,
       mob.maxHealth,
       mob.position,
-      mob.speed,
+      // mob.speed,
       mob.attributes
     );
     this.scene = scene;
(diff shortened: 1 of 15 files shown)


packages/client/src/sprite/sprite_mob.ts
```diff
@@ -46,7 +47,7 @@ export class SpriteMob extends Mob {
       mob.type,
       mob.maxHealth,
       mob.position,
-      mob.speed,
+      // mob.speed,
```
StarDylan, line 50: Instead of commenting out can you delete it? Since it's now an attribute the comment serves no purpose. 

Do that in the other places as well please.

jacksonisidor (approved):
Deleted the lines where we commented out speed. No use for it anymore after putting it in attributes (with health and gold)

packages/client/src/sprite/sprite_mob.ts
jacksonisidor: Right - they have all been deleted now.

PR #110, opened 2025-01-15
fix: mob NPC text color

halliec6:
Fix: Wall Can be Built From Partial Wall 

## Summary:
Fixed a bug where the "doingText" of the mob was not visible on yellow background. 

image

## Changes:

`C:\Users\halli\369\potions\packages\client\src\sprite\sprite_mob.ts` : 

- Changed color of text to black, increased font size from 10px to 12 px 
- added white border to doing text

### Daytime: 
image

### Nighttime:
image

### DIFF
--- a/packages/client/src/sprite/sprite_mob.ts
+++ b/packages/client/src/sprite/sprite_mob.ts
@@ -117,7 +117,7 @@ export class SpriteMob extends Mob {
           fontStyle: 'bold',
           color: '#000000',
           strokeThickness: 2,
-          stroke: '#FFFFFF'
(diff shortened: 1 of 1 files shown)


0felicia0 (approved):

Parshana007 (approved):

StarDylan:
Are there any darker backgrounds that this might be hard to see against? Maybe at night?

Maybe we could outline the black text with a white border, like the name? That way its visible against lots of backdrops

StarDylan (approved):
Awesome, looks great!

PR #105, opened 2025-01-14
Bug fix fence no health

TylerNP:
Group Members: Tyler Pham, Ethan Vos, Taran Singh
#Bug: Fence displays no health bar despite losing health
When the player uses the smash interaction on a fence or a wall, no health bar appears causing confusion to the player regarding the destructibility of the fence or wall.

#Changes
- Added optional instance variables for healthBar, health, and maxHealth in the SpriteItem class in packages/client/src/sprite/sprite_item.ts 
- Updated the constructor for SpriteItem to include initialization of the healthBar, health, and maxHealth for items that are of the 'fence' and 'wall' type
- Added new class method in SpriteItem, updateHealthBar(), to update the displayed health bar based on the percentage of health remaining if the item has a health bar
- Updated the destructor for SpriteItem to remove the healthBar graphic once the SpriteItem is destroyed

#Tests  
/packages/client/test/healthBar/fenceHealthLogic.test.ts
-Unit tests the new logic for SpriteItems using a stubbed out version
- Test constructor to ensure only fence and wall items get a health bar
- Test logic methods isBelowMaxHealth() to ensure health bar only shows up below max health
- Test updateHealthBar() does not create or remove health bars for SpriteItems

PR #114, opened 2025-01-15
fix: don't display health bar for all items

ruiyichi:
**Group:** Daniel Ching, Josh Queja, Keith Lee

### Bug
When starting the game, we noticed items had health bars. Logs and blueberries are items that should not show health bars. 

### Changes
https://github.com/sloalchemist/potions/commit/1c9adc7729b80734e4a4ff4642f87c259901fcd4

- Removes adding the health bar graphic to every item unconditionally

### Testing
- UI change only, no unit test needed


StarDylan (approved):
Good catch! This seems like an oversight from #105.

Issue #51, opened 2025-01-09
Bug: User can walk over edges of water bodies - (Sreshta, Aditya)


sreshtalluri:
Whenever the user double-clicks to get to a point that is on the other side of a water body (i.e. the one in the bottom left of the world), they can cut through the edges of the water tiles.

Here is what the issue looks like:
<img width="175" alt="Screenshot 2025-01-14 at 6 19 36 PM" src="https://github.com/user-attachments/assets/6b5778d4-db49-46a2-a0b7-bcf430eb6a59" />

Issue #112, opened 2025-01-15
Mobs Walk Over Water Bodies


amanikonda123:
During some of the paths that mobs take, they seem to walk on air above the bodies of water. This happens at the top of water bodies as well as some of the sides.

Below are some images showing the bug:
![image](https://github.com/user-attachments/assets/1718a1d4-7662-408a-b16d-9801925c1150)
![image](https://github.com/user-attachments/assets/dcbd5a93-cf89-4e94-b0ea-6a0e0c712972)

Deliverables:
- Mobs should not walk over water bodies

PR #115, opened 2025-01-15
Fix: Bug - User can walk over edges of water bodies

sreshtalluri:
## **Problem:**

Whenever the user double-clicks to get to a point that is on the other side of an unwalkable area, they cut through the edges of the unwalkable area/tiles. This mainly occurs when near the top of the unwalkable areas.

IMG_7971

We found that the issue arises because the pathing algorithm floor of the current point as the start (even if it is in between 2 tiles). The issue occurs when the second click happens as the current point is in between 2 tiles, so it considers the start as the floor, which makes it skip one of the points along its intended path.

## **Solution:**

- Added a ceiling method for coordinates
- Added a check to see if the floored start coordinate has an unwalkable area next to it (depending on the location of the target)
- Using that check, the starting coordinate is chosen as the ceiling or the floor in order to ensure points along the path are not skipped

## **Tests:**

Added few test cases in the pathFinder.test.ts file, to ensure that the path does not skip one of the points along the path. Tested cases where the starting position is in between two tiles and added checks to make sure that the corner coordinate that should be in the path is still there.

**Fixes** [#51](https://github.com/sloalchemist/potions/issues/51)

### DIFF
--- a/packages/common/src/pathFinder.ts
+++ b/packages/common/src/pathFinder.ts
@@ -207,8 +207,20 @@ export class PathFinder {
     return simplifiedPath;
   }
 
-  // Check if surrounding tiles are walkable in specific cases
-  private checkSurrounding(start: Coord, end: Coord): boolean {
+  
+/**
+ * Determines if the surrounding tiles relative to the starting coordinate are walkable.
+ * 
+ * This function checks two adjacent tiles based on the direction from the start to the end
+ * coordinate, verifying if they are walkable.
+ * 
+ * @param start - The starting coordinate.
+ * @param end - The ending coordinate.
+ * 
+ * @returns True if the surrounding tiles in the direction to the end coordinate are walkable, 
+ * false otherwise.
+ */
+  private isSurroundingWalkable(start: Coord, end: Coord): boolean {
     start = floor(start);
 
     if (end.x <= start.x && end.y <= start.y) {
@@ -239,7 +251,7 @@ export class PathFinder {
   ): Coord[] {
     end = floor(end);
 
-    if (this.checkSurrounding(start, end)) {
+    if (this.isSurroundingWalkable(start, end)) {
       start = ceiling(start);
     }
     else {
--- a/packages/common/src/pathFinder.ts
+++ b/packages/common/src/pathFinder.ts
@@ -1,4 +1,4 @@
-import { Coord, floor, step } from './coord';
+import { Coord, floor, round, step } from './coord';
 import { TerrainType } from './terrainType';
 
 export class PathFinder {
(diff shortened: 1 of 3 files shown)


amanikonda123:
Looks good to me. I just want to mention that we saw similar behavior to the other mobs as well, but the fixes we made to the pathing algorithm fixed only the user behavior and not the behavior we saw with the other mobs. We made a new bug (#112) and will address this in a future PR.

packages/common/src/pathFinder.ts
```diff
@@ -207,14 +207,44 @@ export class PathFinder {
     return simplifiedPath;
   }
 
+  // Check if surrounding tiles are walkable in specific cases
```
StarDylan, line 210: `checkSurrounding` is a bit vague, maybe we can rename it to `isSurroundingWalkable` or something more descriptive? 

I also recommend using a JSDoc explaining more of what this function is for like this:
```
/**
 * Add 2 numbers
 * @param a The first number
 * @param b The second number
 */
  return a + b;
}
```

Then we get nice documentation when hovering over it:

[image](https://private-user-images.githubusercontent.com/6405858/402194794-d5f92078-11a3-4fa5-ab00-609999695e20.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY5NTk1MDgsIm5iZiI6MTczNjk1OTIwOCwicGF0aCI6Ii82NDA1ODU4LzQwMjE5NDc5NC1kNWY5MjA3OC0xMWEzLTRmYTUtYWIwMC02MDk5OTk2OTVlMjAucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTVUMTY0MDA4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZDQ0N2Y1M2E1ZGM4MzU2MDVlOTA0NGI0YzA4NDAxMmJkNmM1MjU5ZWExYTJlZDBiYjJlZGU5ZjQ0OTc2ZTEyZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.wiTEFlui_U077yXLpl_sG6wawVx8rnevgYz2UQnf_II)

Here's a nice paragraph from Google describing the difference between line comments and JSdocs [https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments](https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments)

StarDylan (changes requested):

packages/common/src/pathFinder.ts
sreshtalluri: Made the change :)

PR #118, opened 2025-01-15
Fix: resolved openable, smashable bug (Gabe Riedel, Calahan Lackovic)

gabrielriedel:
# Problem

When approaching a smashable item that is walkable and has a layout_type of "opens" (i.e. a gate or a door), there is no button to smash that item.

Screenshot 2025-01-12 at 8 23 25 PM

This issue arose because any items that were walkable but not able to be stood on top of (i.e. a closed gate or door) were getting filtered out from having a smash interaction. 

# Solution

1. Added logic for items who have a layout_type of "opens" to be checked if they were in proximity to the player's position
2. Now all nearby, openable items were added to the array of items that could be interacted with

# Tests

Added one test case for gates and one for doors that checked if the respective object was interactable from a player location one tile away, and it checked if one of the interactions with this object was that it could be smashed.

### DIFF
--- a/packages/client/src/world/controller.ts
+++ b/packages/client/src/world/controller.ts
@@ -236,16 +236,22 @@ function getItemsAtPosition(physicals: Item[], position: Coord): Item[] {
   });
 }
 
-function getInteractablePhysicals(physicals: Item[], playerPos: Coord): Item[] {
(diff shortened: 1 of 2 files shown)


StarDylan (approved):
Looks good! Also just a friendly remember to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format.

PR #127, opened 2025-01-15
Update smashItem Method to Handle Attack Damage

EthanV1920:
## Summary
This PR updates the smashItem method to improve the calculation of attack damage when interacting with a Mob object. The changes ensure that the attack damage is always at least 1, preventing scenarios where the damage could be zero.

## Changes Made
- Store Mob: The mob parameter is stored in the class property for future reference.
- Attack Damage Calculation:
The attack damage is calculated using Math.floor(Math.random() * mob.attack).
If the calculated damage is zero, it defaults to 1 to ensure that the player always inflicts some damage.


StarDylan (changes requested):
Is it possible for you to write a test for this?

StarDylan:
So the issue was: since your code is based on a random function, you weren't sure of a great way to test it.

Hmm, maybe you could add a parameter to the function call that allows you to explicitly set the value in tests?

Like this:  
```ts
smashItem(mob: Mob, rng: () => number = Math.random)
```

Let me know if that rng parameter doesn't make sense

packages/server/test/items/itemSmashable.test.ts
```diff
@@ -0,0 +1,137 @@
...
+//             halfHealthFence.updateHealthBar();
+//             expect(halfHealthFence.healthBar).toBeDefined();
+
+//             const fullHealthGate = new SpriteItem(maxHealth, maxHealth, mockScene, { layoutType: 'gate' });
+//             expect(fullHealthGate.healthBar).not.toBeDefined();
+//             fullHealthGate.updateHealthBar();
+//             expect(fullHealthGate.healthBar).not.toBeDefined();
+
+//         });
+//     });
```
StarDylan, line 136: I would remove this section, since commented out areas just add noise to the codebase.

packages/server/src/items/smashable.ts
```diff
@@ -19,9 +19,12 @@ export class Smashable {
 
   private mob!: Mob; // Add mob as a class property
 
-  smashItem(mob: Mob) {
+  // default param for testability, only call with one param in normal use
```
StarDylan, line 22: I recommend using a JSDoc for this, like so:
```
/**
 * Add 2 numbers
 * @param a The first number
 * @param b The second number
 */
  return a + b;
}
```

Then we get nice documentation when hovering over it.

Here's a nice paragraph from Google describing the difference between line comments and JSdocs [https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments](https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments)

StarDylan (changes requested):

StarDylan (approved):
Looks great!

PR #135, opened 2025-01-17
Fix item spawning

jmdebarro:
Created new func in item.ts
Edited itemSpawning.ts to account for new local max
Edited config files so spawn_item actions had new parameters (global_max, local_max, radius)
Created new test file

### DIFF
--- a/packages/server/src/items/on_ticks/spawnItem.ts
+++ b/packages/server/src/items/on_ticks/spawnItem.ts
@@ -14,11 +14,16 @@ export class SpawnItem implements OnTick {
     }
 
     const itemType = parameters['type'] as string;
-    const count = Item.countTypeOfItem(itemType);
-    const max = parameters['max'] as number;
+    const radius = parameters['radius'] as number;
+    const globalCount = Item.countTypeOfItem(itemType);
+    const localCount = Item.countTypeOfItemInRadius(itemType, item.position, radius);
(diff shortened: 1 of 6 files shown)


logancostello (approved):
Looks good to me. Make sure we squash the commits properly on merge.

packages/server/src/items/on_ticks/spawnItem.ts
```diff
@@ -14,11 +14,16 @@ export class SpawnItem implements OnTick {
     }
 
     const itemType = parameters['type'] as string;
-    const count = Item.countTypeOfItem(itemType);
-    const max = parameters['max'] as number;
+    const radius = parameters['radius'] as number;
+    const globalCount = Item.countTypeOfItem(itemType);
+    const localCount = Item.countTypeOfItemInRadius(itemType, item.position, radius);
+    Item.countTypeOfItemInRadius(itemType, item.position, radius);
```
StarDylan, line 20: This is seemingly just calling the function without saving the return value?

StarDylan (changes requested):

packages/server/src/items/on_ticks/spawnItem.ts
logancostello: You're right. It was previously commented out when testing manually. Will remove now.

StarDylan (approved):
Awesome looks great! Love the conventional commits!

Issue #116, opened 2025-01-15
Bug: Clicking on an unwalkable tile causes user to stop in the middle of their path (Aditya, Sreshta)


amanikonda123:
In general, when trying to move to an unwalkable tile, the user does not move anywhere. So, the fix to this would be to get to the closest walkable location/tile to the point that they originally clicked.

Another issue that stems from this is that when the user is already moving along a path and then clicks on an unwalkable tile, the user character just stops in place (no matter where it is, even if in between tiles; the character just freezes and it doesn't look smooth). So, we want to eliminate this by allowing the user to walk as close as they can to the unwalkable tile.

PR #143, opened 2025-01-17
Fix: Bug - Clicking on an unwalkable tile causes user to stop in the middle of their path

amanikonda123:
## **Problem:**

In general, when trying to move to an unwalkable tile, the user does not move anywhere. So, the fix to this would be to get to the closest walkable location/tile to the point that they originally clicked.

Another issue that stems from this is that when the user is already moving along a path and then clicks on an unwalkable tile, the user character just stops in place (no matter where it is, even if in between tiles; the character simply freezes and feels awkward on the user side. So, we want to eliminate this by allowing the user to walk as close as they can to the unwalkable tile.

## **Solution:**

- Implemented a new function called findNearestWalkableTile() that performs a breadth-first search from the unwalkable end coordinate by checking all 8 of its adjacent tiles, and finds the first walkable tile. If no such tile can be found, then it throws an error
- generatePath() calls findNearestWalkableTile() to replace the end coordinate if it's unwalkable. If no walkable tile can be found, then an empty path is returned, which was the default behavior of the function prior to our modifications.

## **Tests:**

- Added 2 test cases in the pathFinder.test.ts file to test the behavior of generatePath(), ensuring that the user walks to the closest walkable tile if the original end point is unwalkable
- To test for coverage, we added another test case if there is no walkable tile the user could move to. The expected default behavior is that the user stays in place.

**Fixes** [#116 ](https://github.com/sloalchemist/potions/issues/116)

### DIFF
--- a/packages/common/test/pathFinder.test.ts
+++ b/packages/common/test/pathFinder.test.ts
@@ -91,15 +91,6 @@ describe('PathFinder', () => {
     expect(path).toContainEqual(end);
   });
 
-  // FIX: This test is no longer needed since we want to walk to the closest walkable tile if the goal is not walkable
-  //
-  // test('generatePath should return an empty path if the goal is not walkable', () => {
-  //   const start = { x: 0, y: 0 };
-  //   const end = { x: 1, y: 1 }; // Blocked
-  //   const path = pathFinder.generatePath([], start, end, false);
-  //   expect(path).toEqual([]);
-  // });
-
   test('generatePath should not cut over unwalkable corners (#1)', () => {
     // check not cutting top left corner
     const start = { x: 0.75, y: 0 };
--- a/packages/common/test/pathFinder.test.ts
+++ b/packages/common/test/pathFinder.test.ts
@@ -3,6 +3,7 @@ import { PathFinder } from '../src/pathFinder';
 describe('PathFinder', () => {
   let pathFinder: PathFinder;
   let pathFinder2: PathFinder;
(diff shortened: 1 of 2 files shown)


packages/common/test/pathFinder.test.ts
```diff
@@ -75,12 +91,14 @@ describe('PathFinder', () => {
...
-    expect(path).toEqual([]);
-  });
+  // FIX: This test is no longer needed since we want to walk to the closest walkable tile if the goal is not walkable
+  //
+  // test('generatePath should return an empty path if the goal is not walkable', () => {
+  //   const start = { x: 0, y: 0 };
+  //   const end = { x: 1, y: 1 }; // Blocked
+  //   const path = pathFinder.generatePath([], start, end, false);
+  //   expect(path).toEqual([]);
+  // });
```
StarDylan, line 101: In general, I prefer to remove code and not comment it out. Since it adds noise to the codebase and can pile up over time. I think it would be better if we removed it.

StarDylan (changes requested):

packages/common/test/pathFinder.test.ts
sreshtalluri: Sounds good :)

sreshtalluri (commented):
I think everything should be good now that we deleted the commented out / unnecessary test case.

StarDylan (approved):
Looks great! Love the conventional commits!

PR #144, opened 2025-01-18
fix: interact buttons persist at long distance

gabrielriedel:
Changes:

- Abstracted logic of 'push' and clearing buttons for both interact and chat buttons to ButtonManager class
- interactButtons now stored as a ButtonManger object, so the proper clear function is called to clear interactions
- Wrote unit tests to confirm that both chat and interact buttons were being properly cleared by clearButtonOptions().

Aside:
We originally were just trying to fix the clearing of interact buttons, as chat buttons were already "managed" by a ChatButtonManager class. So we created an InteractButtonManager class to handle similar functionality as the ChatButtonManager. However, we realized afterwards that it would be more efficient and abstract to have a single ButtonManager class that managed both interact and chat buttons. We recognize that the re-factorization is slightly breaking the Atomic Pull Requests contribution standard, but we have thoroughly tested both cases of interact and chat buttons to make sure their clearing still functions properly.


StarDylan (approved):
Looks great!

Issue #104, opened 2025-01-14
Bug: Gate and Door has no healthbar


DarrenH6293:
Fences show health bars when being damage, but doors and gates do not, showing that healthbar appearance is inconsistent across objects.

PR #145, opened 2025-01-18
fix: doors and gates have no healthbar

Andrew76xx:

**Bug Description** (Issue #104): When smashing the gate fence or the door, the health bar will not show. 

**What Changed**: We adjusted the if statement in `sprite_item.ts` and added `layout_type === 'open'`, representing both the gate and the door, allowing the health bar to be present when the player presses the smash command.

**Why**: This addition of the health bars allows the player to know the progress of getting rid of the door and gate. This also removes confusion on whether or not the gate and the door can be broken. It makes showing health bars consistent since fences also have this feature.

Before:
image

After:
image

### DIFF
--- a/packages/client/src/sprite/sprite_item.ts
+++ b/packages/client/src/sprite/sprite_item.ts
@@ -34,7 +34,7 @@ export class SpriteItem extends Item {
       this.attributes[key] = item.attributes[key];
     }
 
-    if(this.itemType.layout_type === 'fence' || this.itemType.layout_type === 'wall' || this.itemType.layout_type === 'opens') {
+    if(this.maxHealth) {
       this.healthBar = scene.add.graphics();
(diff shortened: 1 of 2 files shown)


StarDylan (changes requested):
Is there a way to replace these if statements with something that applies to everything needing health? We should be able to make new items with just changing the config.

Adding explicit if statements makes it way more difficult if we wanted to add a new item that needs a health bar.

Andrew76xx:
> Is there a way to replace these if statements with something that applies to everything needing health? We should be able to make new items with just changing the config.
> 
> Adding explicit if statements makes it way more difficult if we wanted to add a new item that needs a health bar.

Andrew76xx:
image

Yes. @StarDylan

StarDylan (approved):
Looks good!

PR #146, opened 2025-01-18
Return to start/load game screen after death

ethanhandelman:
Group: Andrew Estrada, Ethan Handelman

**Bug Description**
After the player's health is reduced to zero causing death, the game-over screen appears but there is no way to respawn in-game without refreshing the page through the browser.

**Changes Made**
Created function resetToLoadWorldScene() which waits three seconds before stopping all scenes that are currently active and restarting the LoadWorldScene. This effectively returns the client to the start of the game.
Added RESPAWN_DELAY constant to set delay for start game scene to be shown (in ms)
`packages/client/src/scenes/worldScene.ts`

Added call to resetToLoadWorldScene() in handling of player death
`packages/client/src/services/serverToBroadcast.ts`

Added init() function to reset lastAnimationKey to blank, ensuring proper rendering of character on load screen
Added call to events to prevent duplicate key creation on revive
`packages/client/src/scenes/loadWorldScene.ts`

_Note: Not fit for unit tests as discussed, integration testing needed_

### DIFF
--- a/packages/client/src/scenes/loadWorldScene.ts
+++ b/packages/client/src/scenes/loadWorldScene.ts
@@ -34,6 +34,15 @@ export class LoadWorldScene extends Phaser.Scene {
   paletteSwapper: PaletteSwapper = PaletteSwapper.getInstance();
   lastAnimationKey: string = '';
 
+  /* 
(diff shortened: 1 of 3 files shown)


StarDylan (approved):
Looks great!

PR #148, opened 2025-01-19
Bug Fix: Only 1 interaction when the player is on top of a non-walkable object

spruhaN:
Previously when the player would place a partial wall(non-walkable object) it would register twice:

- once as a nearby object
- once as an object the player is on top

We now are enforcing that the list of objects that interactions are based on is unique so it will only prompt one interaction.


annamakarew (approved):

sbryukho (approved):

StarDylan (approved):
Looks good!

PR #150, opened 2025-01-20
Bug Fix Cannot Walk Through Open Gates

MorrowCameron:
Problem Statement:
Currently player cannot move through gate, even when it is open.

Solution Overview:
Added attribute to openable item type to check if item has been opened during pathing. Open gates shouldn't block pathing. 

Testing:
Created a new unit test and manually tested


zimmerjonathan (approved):
LGTM

irwinj22:
LGTM also

StarDylan (approved):
Looks good!

PR #157, opened 2025-01-21
Fix: Issue 96 - Prettier Formatting Issues Are Not Caught by Linter (Ryan Hu, Ryan Chan, Kyle Taschek)

### LINKED ISSUE #96: Bug: Prettier formatting issues are not caught by linter (Ryan Hu, Kyle Taschek, Ryan Chan)

ktaschek:
Closes #96

## Summary

This PR resolves the issue of the linter not formatting files properly by installing the Prettier plugin for Eslint.

## Changes

`eslint.config.mjs`
- imported Prettier Plugin for eslint

`package.json`
- added lint fix script: `lint:fix`

All other files:
- fixed linting issues

## Testing
- Ran linter script to ensure functionality
- Built and ran locally


package.json
```diff
@@ -5,12 +5,15 @@
   "scripts": {
     "format": "prettier --config .prettierrc '**/*.ts' --write",
```
StarDylan, line 6: It seems the single quotes don't work on Windows, when I replace it with `**/*.ts` instead of `'**/*.ts'` it works.

StarDylan (changes requested):
Is there a reason why we are adding prettier rules to eslint vs. keeping them two separate tools? 

Here's an article which highlights some pros vs cons: https://prettier.io/docs/en/integrating-with-linters.html. Otherwise we could just use [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to make sure they don't conflict

ryanhu021:
@StarDylan thanks for sending over the resources! We considered keeping Prettier and ESLint separate when approaching this issue, but ultimately decided on combining them for several reasons:

1. The biggest issue with the original configuration was that code editors (such as VSCode and IntelliJ) made it challenging to enforce Prettier and ESLint rules simultaneously. These editors typically allow you to select a default formatting extension, which in this case can either be Prettier **OR** ESLint, leading to one set of rules being auto-enforced by the editor and the other requiring a manual `npm` command. Either way, code contributors will have to run `pnpm format` or `pnpm lint:fix` every time they make a commit, creating an extra step for contributors to remember. 
2. The issue mentioned above also increases the risk of inconsistencies that are not caught until the pull request stage. Many style and quality inconsistencies could slip through the review process, especially less noticeable issues like indenting, spacing, and quote types. This makes it challenging to maintain a clean and extensible codebase since there is a higher chance of each contributor's change creating merge conflicts on small code inconsistencies. From our experience working on issues before this, it was especially noticeable in unit tests, where many test files had drastically different formats when we wanted to reference them.
3. The Prettier documentation states that _"You end up with a lot of red squiggly lines in your editor, which gets annoying."_ While this is true, it is better to catch these error in local development instead of the review stage, especially for a codebase that has so many contributors. Code editors make this easy with formatting options that auto-resolve these errors. Combining Prettier and ESLint into a single command reduces the complexity of making a contribution since it moves more of the formatting responsibility from the contributor to the code editor, which increases the likelihood of these style/quality issues being caught at the local development stage.
4. The documentation page also states that Prettier ESLint plugins _"are slower than running Prettier directly."_ From a contributor's perspective, while running the ESLint command itself may be a little slower, reducing the number of commands we have to run from two (`pnpm format` and `pnpm lint:fix`) to one (`pnpm lint:fix`) still reduces the overall contribution time (since we no longer have to run `pnpm format`). We can instead run one command to fix all formatting and styling issues.
5. _"They’re yet one layer of indirection where things may break."_ While this is also true, the `eslint-plugin-prettier` is a mature npm package that has >8000 dependents on npm and is used in numerous commercial applications.

StarDylan:
@ryanhu021  Thanks for the response! Do you think we should remove the `format` command then? So we just have to run that one command? We might also need to update the CI to reflect us moving everything to the one command.

ryanhu021:
@StarDylan yes that makes sense. We'll remove the `format` command from `package.json` and the CI script.

ryanhu021:
@StarDylan should be good now!

StarDylan (approved):
Looks good!

ryanhu021:
Doing one more rebase to get latest changes from main.

PR #159, opened 2025-01-21
Feat: Adding a timer to blue potion speed effect

eswenke:
## Summary
- Added a 15 second timer to the speed on the blue potion. Previously stacking was possible up to 10 speed, but a timer makes more sense and timer logic can be improved upon and later used in other cases.

## Changes
- `packages/server/src/items/potionEffects.ts`  Added a specific check for blue potions, set up an interval checking for speed reset.
- `packages/server/src/mobs/mob.ts` Added target_speed_tick as a property and parameter. Added a getter along with a current tick getter. Also added a changeTargetSpeedTick function as a setter. Ended up not using it, but could be useful to have in the future for tick related effects. Added a function to check if time has expired and then changed speed accordingly. Changed mob table query to include target_speed_tick, and made sure to collect that in other queries.
- `packages/server/src/services/clientCommunication/pubsub.ts`  Added 'changeTargetSpeedTick' signature.
- `packages/server/src/services/clientCommunication/stubbedPubSub.ts` Added 'changeTargetSpeedTick' signature.
- `packages/server/src/services/clientCommunication/ablyService.ts` Added 'changeTargetSpeedTick' method for broadcasting mobChange.
- `packages/server/src/services/clientCommunication/clientMarshalling.ts` Added target_speed_tick as an attribute to the player rather than a regular property.
- `packages/client/src/world/sprite_mob.ts` Edit ux to hide tick changes. Changed the additional speed effect to green and the slowdown to red.
- `packages/client/src/worldMetadata.ts` Added target_speed_tick getter. Not used but could be helpful in future.
- `packages/client/src/world/controller.ts` Added real time update on the speed attribute in game (info tab).
- `packages/server/test/items/potionEffects.test.ts` Edited unit tests to include tick checks.

### DIFF
--- a/packages/server/src/items/potionEffects.ts
+++ b/packages/server/src/items/potionEffects.ts
@@ -1,7 +1,5 @@
 import { Mob } from '../mobs/mob';
-import { hexStringToNumber, numberToHexString } from '../util/colorUtil';
-import { FantasyDate } from '../../src/date/fantasyDate';
-import { scheduler } from 'timers/promises';
+import { hexStringToNumber } from '../util/colorUtil';
 
 
 export function drinkPotion(mob: Mob, potionType: string): boolean {
--- a/packages/server/src/items/potionEffects.ts
+++ b/packages/server/src/items/potionEffects.ts
@@ -10,23 +10,23 @@ export function drinkPotion(mob: Mob, potionType: string): boolean {
     mob.changeHealth(50);
     return true;
   }
+
   else if (potionType === String(hexStringToNumber('#0000FF'))) {
-    // get current tick, add x amount of time, check when global tick hits that number, change speed
-    let curr_date = FantasyDate.currentDate();
-    const target_tick = curr_date.global_tick + 10;
-    console.log("target tick:" + target_tick.toString());
 
-    console.log("initial tick:")
-    console.log(curr_date.global_tick);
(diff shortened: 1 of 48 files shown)


packages/server/src/items/potionEffects.ts
```diff
@@ -8,6 +8,25 @@ export function drinkPotion(mob: Mob, potionType: string): boolean {
...
+    mob.changeSpeed(speedDelta, speedDuration);
+
+    // Start checking for speed reset periodically
+    const resetInterval = setInterval(() => {
+      if (mob.checkSpeedReset(speedDelta)) {
+        clearInterval(resetInterval);  // Stop checking once the speed is reset
+      }
+    }, 500);
+
+    resetInterval.unref(); // allows teardown if last active handler
```
StarDylan, line 26: My only concern about this is if the server resets, then this will never be called.

StarDylan (changes requested):

packages/server/src/items/potionEffects.ts
jacksonisidor: Implemented a different solution for this using 'tick' in mob.ts that calls the function every tick. Also merged again with the new changes. @StarDylan

StarDylan (approved):
Looks good!

PR #164, opened 2025-01-21
feat: smashing fence/partial wall drops log

emlai1263:
# Issue 43 (Emily Lai, Adrian Elias)

# Summary 
- Partial walls, fences, and walls now drop logs upon being smashed

# Changes
- itemDropConfig mapping is created in smashable.ts to indicate what should be dropped when a certain item is destroyed
- added corresponding tests in fenceDropTest to for partial walls, fences, and walls
- Refactored the Smashable logic by abstracting item drop behavior into a configurable structure, enabling flexibility for extending similar functionality to other entity types


Twicee:
It should be just partial walls and fences. Walls already had the functionality to drop logs.

packages/server/src/items/smashable.ts
```diff
@@ -27,28 +27,44 @@ export class Smashable {
...
-  destroyPotionStand(){
+  // destroying a fence or wall should return the log use to build it
+  dropLog(item: Item) {
+    itemGenerator.createItem({
+      type: 'log',
+      position: item.position
+    });
+  }
+
+  destroyPotionStand() {
```
StarDylan, line 56: Instead of hard coding this logic in the `smashable.ts`, can we abstract it away so we can apply similar drops to other entities as well? It probably should be configurable in the config

StarDylan (changes requested):

packages/server/src/items/smashable.ts
emlai1263: We could try making a new type for these types of destructible walls, do you think that would work? Also Where do you mean by the config?

StarDylan: So in `/packages/server/data/global.json` and `/packages/client/static/global.json` we define how the world works. Ideally that is where we define all the behavior for an item. Otherwise I'm not sure if creating a new type is the best, since we probably want all walls to be destructible? But feel free to argue otherwise

emlai1263: Last night we created a mapping in item.ts that lets you specify the item (fence, wall, etc) and what it drops (logs). Is that alright?

emlai1263: We spoke with the prof, we'll continue with this abstraction and implement it into a bigger feature this week

StarDylan: We can continue it here

StarDylan (commented):
Waiting on merge for the refactor next week (correct me if I'm wrong)

StarDylan:
See #164 for some related work

Twicee:
@StarDylan Sorry for the confusion but it looks like we will be just abstracting the drop items. No more to come as previously mentioned.

StarDylan (approved):
Ok, it Looks good!

StarDylan:
Also friendly reminder to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format for future commits.

PR #169, opened 2025-01-21
154 refactor potion retrieval from stand

chelseylearns:
### Group Members: Alex Ralston, Harshini Karthikeyan, Chelsey Harper, Bernette Chan

Fixes [#154](https://github.com/sloalchemist/potions/issues/154) **(Refactor)**

### Summary

Previously we'd fixed the bug where players cannot pick up their own potions from potion stands without purchasing them. To do this, we added a retrieve_item method to the Purchaseable class to transfer the item from the stand to the mob. 

Now, we roll back this change and fix the issue another way. We make a potion stand an instance of Container, so that we can use built-in Container logic to retrieve potions from the stand. This is a cleaner fix. 

### What Changed

-  `packages/server/test/items/itemRetrievePotionFromStand.test.ts`: Refactored to use container's getItem instead of Retrieve.

-  `packages/server/src/items/uses/usesConfig.ts`: Removed retrieve entry.

-  `packages/server/src/items/uses/stand/retrieve.ts`: Deleted.

-  `packages/server/src/items/purchasable.ts`: Removed retrieveItem method.

- `packages/server/src/items/container.ts`: Added item subtype to itemGenerator's createItem.

- `packages/client/static/global.json`: Changed potion-stand's interaction retrieval to use get_item instead of old retrieve_item.


StarDylan (approved):
Looks good! It simplifies the logic quite a bit

PR #181, opened 2025-01-22
Refactor relevant variables with `as const` for enhanced type ergonomics

AdamMikacich:
**Group**: Adam Mikacich, Ananya Thapar, Caroline Koddenberg

**Refactor description**:
Prior to this PR, multiple immutable arrays (aka tuples) and other objects were not typed with the `as const` annotation. This annotation provides benefits to the developer by telling the language server that the contents with these objects will not change (note this is different from the `const` keyword which only affects the variable reference, not the content of the objects themselves). This code hygiene improvement ensures the tuple or object will not be accidentally modified elsewhere. Moreover, using `as const` with a `satisfies` constraint ensures the types within the object are as expected (i.e. only a tuple of strings, not strings & numbers). This new implementation also has the benefit of automatic type narrowing when accessing values in a tuple or object (please view the examples below).

**What changed**:
For all relevant constant objects (including tuples and Record types), we added the correct `as constant satisfies...` annotation to tell TypeScript to restrict the type and ensure it is not modified. This enhances type inference within the codebase by allowing TypeScript to display a more specific type since it will be known at compile time (i.e. displaying the type of `englishPrefixes[0]` as the string literal `'Ash'` instead of a generic `string`). Note we also had to remove the type declaration on the variable itself to prevent the type from being generalized again (i.e. back to `string[]` instead of `['Ash', ...]`)

**Why**:
Improved code ergonomics and type narrowing as explained above and demonstrated below.

**Before**:
Screenshot 2025-01-22 at 3 37 15 PM

Screenshot 2025-01-22 at 3 33 16 PM

**After**:
Screenshot 2025-01-22 at 3 37 35 PM

Screenshot 2025-01-22 at 3 32 54 PM


StarDylan (approved):
Looks great, love the extra type safety

PR #185, opened 2025-01-24
Smarter NPCs

logancostello:
**Large bug**: Mobs do not work together, despite being in the same alliance with the same goals. For example, multiple mobs will try to gather the same gold on the ground despite only one of them being required for this task. They should be smarter and work as an team

**Solution**: When calculating the findItem cost, a penalty (Infinity) will be applied if other mobs in the same alliance are already going for that item. Now mobs look for more than one item incase the item they initially want is targeted. This small change has a huge impact, as mobs now appear to be much smarter. You can see the change in this video, where 3 mobs collect blueberries, but they all target different blueberries on the ground.

https://github.com/user-attachments/assets/4ae7b7d2-cc74-4482-9a3e-03220bb37180

We still need to write unit tests, but this is currently a draft PR anyways.


StarDylan (approved):
Looks good to me!

jmdebarro (approved):
lgtm, changed test file name.

AndrewJMart (approved):
The Don Approves

PR #197, opened 2025-01-27
Enhancement: VSCode Task Script to start application

annamakarew:
Previously, a user would have to run a series of commands in different terminals(in a specific order) to build the client.

By adding a VSCode Task this can all be streamlined into just triggering the specific task 'Start All', which spins up the auth server, server, and client into the task system.

One could also trigger specific components of the build like 'Client', 'Auth Server', and 'Server' and it will accordingly run the depending tasks.


StarDylan (approved):
Looks good!

PR #203, opened 2025-01-27
Feat: Create Developer Shortcuts to increase health and speed

Parshana007:
Feat: Create Developer Shortcuts to increase health and speed

#Members: Felicia Patel, Parshana Sekhon, Hallie Christopherson

## Summary:
Pressing Shift+F increases a player's speed by a factor of 2 from the original speed. 
Pressing Shift+H sets a player's health back to full health [100]. 

## Changes:

`packages/client/src/scenes/worldScene.ts` : 
- Added the keyboard shortcuts that trigger client side speed of character and health

`packages/common/src/playerToServer.ts`:
- Added cheat: { action: string } to the message map of playerToServer where action currently includes "health" and "speed" was left as a string to be more inclusive of more actions in the future. 

`packages/server/src/services/clientCommunication/ablyService.ts`:
- Allowing communication between server and client with subscribeToPlayerChannel

`packages/server/src/services/developerCheats.ts`
- added applyCheat function
- setup so that developers can easily add new cheats

`packages/client/src/utils/developerCheats.ts`:
- Broadcasts PlayerToServer message here for health and speed

`packages/client/src/services/playerToServer.ts`:
- Changed publishPlayerMessage so that the function is exported; this was to create the util/developerCheats.ts in the event that more cheats will be created in the future.  

`packages/server/test/developerCheats/developerCheats.test.ts`
- added unit tests to check that applyCheats calls the correct function

`wiki pages`: Developer Guide
- If a developer wants to reset their health to develop and test real-time features they can. For example, if you are testing fence building but a mob type keeps trying to kill the player, resetting the health will ease the developer by not having to return to the restart game page. Furthermore, to improve developer experience, speed was increased from one location to another faster.


halliec6 (approved):

0felicia0 (approved):

StarDylan (approved):
Looks good, in the future maybe we can add some flag to the account that they are a Dev and not a normal player, so other players can't just get infinite health.

PR #204, opened 2025-01-28
fix: players now able to give items to villagers

gabrielriedel:
Fixed logic that prevented players and villagers from giving each other items despite having an alliance. Wrote a tests case and added logic to testSetup to support the test case.


StarDylan (approved):
Looks great!

PR #207, opened 2025-01-29
feat: Support Docker for server package

ruiyichi:
**Group:** Daniel Ching, Josh Queja, Keith Lee

### Description

To eventually support the deployment of the potions server on remote machines, we've added support for the **server** package to run in its own Docker container. Developers have the option to either run the server locally or inside the Docker container.

### Changes
https://github.com/sloalchemist/potions/commit/13b0252bd89ec79c976bf4af1a6e0ac1d7a81bf4

- Created Dockerfile to configure and build the server package in the container
- Updated Continuous Integration Checks to verify Docker builds successfully

### Wiki Page
The Wiki page to setting up and running Docker can be found [here](https://github.com/sloalchemist/potions/wiki/Docker-Guide).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:20-alpine
+
+WORKDIR /usr/src/app/packages
+
+COPY packages/server ./server
+COPY packages/converse ./converse
+COPY packages/common ./common
+
+WORKDIR /usr/src/app
+
+COPY package.json .
+COPY *.yaml .
+COPY tsconfig.json .
+
+RUN npm install -g pnpm
+RUN pnpm install
+RUN pnpm build
+
+WORKDIR /usr/src/app/packages/server
+CMD ["sh", "-c", "pnpm run create test-world && pnpm dev test-world"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY package.json .
+COPY *.yaml .
+COPY tsconfig.json .
+
+RUN npm install -g pnpm
+RUN pnpm install
+RUN pnpm build
+
+WORKDIR /usr/src/app/packages/server
+CMD ["sh", "-c", "pnpm run create test-world && pnpm dev test-world"]
```
StarDylan, line 20: So this Dockerfile is a bit inefficient. Since any code change will cause it to reinstall all the dependencies to rebuild the docker container. I'd recommend first copying the package.json and other config files, install the dependencies, then copy everything else over. 

Here's a resource that describes it well:
https://www.blacksmith.sh/blog/how-to-optimize-dockerfile-faster-docker-builds

StarDylan (changes requested):

Dockerfile
ruiyichi: We asked Professor Pierce if we should implement caching the dependency packages, and he said it was ok without it since devs will likely run the server locally when developing.

Issue #122, opened 2025-01-15
Refactor: Apply Evolvability to createStandFromItem


BrentNakashima:
The only world object that can be created from an item is a potion-stand, so incorporate evolvability in the case that different items can create different world objects.

joshualeeee:
Steps for fix: 
1. Add containerType and canCreateContainer attributes in item class
2. Refactor stand/createStand -> move to container/createContainer
3. Refactor purchasable/createStandFromItem, use createContainer instead of createStand

PR #208, opened 2025-01-29
Refactor create stand from item & Refactoring Prettier . Issue #122

apervmm:
Issue #122

The only world object that can be created from an item is a potion stand, so incorporate evolvability in the case that different items can create different world objects.

Steps for fix:

Add class Create that can be use to generically create objects in the world
Refactor stand/createStand to use create object
Delete purchasable/createStandFromItem


packages/server/src/items/purchasable.ts
```diff
@@ -88,22 +88,22 @@ export class Purchasable {
...
+  //       templateType: createStandFrom.type,
+  //       items: 1,
+  //       capacity: 20
+  //     }
+  //   });
+
+  //   createStandFrom.destroy();
+
+  //   return true;
+  // }
```
StarDylan, line 106: Since we aren't using this code, lets just remove it and not comment it out. It adds to the noise of the codebase if we leave it.

packages/converse/data/knowledge-graph.db-shm
StarDylan, line 1: This should also be removed with
```
git rm --cached packages/converse/data/knowledge-graph.db-shm
```
And added to the .gitignore (if it's not already added)

packages/converse/data/knowledge-graph.db-wal
StarDylan, line 1: This file should also be removed

StarDylan (changes requested):
Some minor things, but looks great otherwise

StarDylan (approved):
Looks great!

PR #218, opened 2025-01-29
fix(client): Increased the depth of health bars to the top

### LINKED ISSUE #172: Bug: Health Bars are covered by most all structures
Health bars are drawn at a lower layer than the graphics for the structures. This is problematic because it hides the health bar of structures and mobs, e.g. , doors which have health bars that are always hidden due to this. This can cause confusion for the user regarding if the door can actually be destroyed.

TylerNP:
Changes
- Add drawing depth for sprite_items and sprite_mobs to the top
- Closes #172


StarDylan (approved):
Looks good!

Issue #221, opened 2025-01-29
Testing Wiki Page, Adding Test Templates, and Highlighting Code Coverage


smbock42:
Added [Testing and Debugging Wiki page](https://github.com/sloalchemist/potions/wiki/Testing-and-Debugging). 

Added Server Testing Templates

Will provide more organized and clean testing while also highlighting current code coverage.

PR #223, opened 2025-01-29
Feature: Create item and mob test templates

smbock42:
Linking issue #221  - Adding item and mob server test templates. This will prevent the need to create boiler plate in the future for server tests.


StarDylan (approved):
Looks good!

PR #237, opened 2025-01-30
Bug fix item health reset

tarannssingh:
Changes:
- remove hardcoded max health
- get max health from world scene if the sprite has a health attribute


TylerNP:
looks ok to me

StarDylan (approved):
Looks good!

PR #247, opened 2025-01-30
Add WASD as movement option

xgarcia17:
**Fix:** In addition to current/original movement by clicking with the mouse, we've added the ability to move using the WASD keys on the keyboard

**Note:** To prevent a denial of service from spamming movement messages to Ably, we have prevented messages from being sent when the same key is pressed more than once before the target is reached. This doesn't interrupt any movement from the player's perspective.

**Files altered:** client\src\scenes\worldScene.ts in create()


StarDylan (changes requested):
> This current implementation is intended to handle individual key presses, although holding down keys does work. However, the issue with holding down keys is that messages are continually sent to Ably, so we advise against this.

Maybe we could add some rate limit? Or disable it until we reach the new location? Something that doesn't let our own players Denial of Service our game.

xgarcia17:
Closed: this issue became a two-week issue which has it's own PR.

PR #249, opened 2025-01-30
Fix: Fixed health bar not displaying over items

AkashPraveen:
Health bars would not display over items that existed such as walls. (You couldn't see door health bars in the village when smashing them). Corrected the graphics so the health bar would show.


StarDylan (approved):
Looks good!

Issue #209, opened 2025-01-29
Update Player Profile


anvch:
The player has information like health, attack, speed, etc. that needs to be shown. These stats may vary based on affiliation/status effects. Make sure the basic layout of this information is set-up.

Deliverables:
* Make sure user/player schema is up to date with fields such as health, affiliation, speed, attack, etc.
* Create default values for (new) players.
* Show this player information in-game (player tab).
** These values are mutable through potions (temp) / affiliations (permanent).
** Do not need to implement status effects, that is a separate issue.

PR #263, opened 2025-02-02
Player profile

Andrew76xx:

**Bug Description** (Issue https://github.com/sloalchemist/potions/issues/209): Making sure the Player Profile can present the attack stats and personality stats, as well as making them mutable for any future features. 

**What Changed**: We changed `mob.ts` in the common folder to allow all MobIs to contain personalities. We added functions in `mob.ts` in the server folder to where the stats can be mutable. In `clientMarshalling`, we formatted the personality data so that the client is able to read it. In `uxScene`, we updated User Interface to contain new attacks and personality stats. The default values of the stats were already provided by the `global.json` in the server folder. We used `community_id` for the players affiliation and is unmutable.

Result:
image


StarDylan (approved):
Looks good!

PR #271, opened 2025-02-05
feat: add player chat

ruiyichi:

**Feature**
Added a mechanism for players to enter freeform chat. The chat box can be toggled with "/" and dismissed with the "ESC" key. When worlds can support multiple players, this mechanism will need to be updated to broadcast the message to all players.

**Changes**
https://github.com/sloalchemist/potions/commit/a182a77d63f8bb178cf47794fc06965bf1aef044

- Create a scene to display the text input and show it

https://github.com/sloalchemist/potions/commit/eca36b0e6ab983c43029a475df181f95f4ff087d

- Fix linter issues

**Testing**
UI change only, no unit test needed


packages/client/src/scenes/chatOverlayScene.ts
```diff
@@ -0,0 +1,92 @@
...
+        const words = message.split(' ');
+        let isValid = true;
+
+        // No word should exceed 10 in length
+        for (const word of words) {
+          if (word.length > 10) {
+            isValid = false;
+            break;
+          }
+        }
```
StarDylan, line 64: Is there a reason why this is imposed? It seems a bit restrictive

StarDylan (changes requested):

packages/client/src/scenes/chatOverlayScene.ts
ruiyichi: This is to ensure the text input doesn't overflow the speech bubble.

jqueja: Before the Change:
image

After the Change:
image

In the sense of making it more restrictive, we wanted to make sure that no one spams the chat. Thus making sure the communication between players was more concise rather than spammy.

StarDylan (approved):
Looks good

PR #284, opened 2025-02-05
Moving wasd to update

kailaaquino:
### Feature: Added keyboard movement with WASD keys. 
Allows for combination of keys for diagonal movement. Does not spam Ably with messages. 

### What changed: 
`packages/client/src/scenes/worldScene.ts `
In create() we are setting booleans for WASD keydown and keyup and saving it to a dictionary.
In update() we are checking the dictionary booleans and incrementing the x and y coordinates according to the keys pressed with handlePlayerMovement. To prevent spamming Ably with requests, we set a counter to publishPlayerPosition every 50 updates, if there has been a change in movement.


StarDylan (approved):
Looks great! Thanks for adding the checks for publishing the movement!

PR #290, opened 2025-02-06
Implementing Server Spin up and Client linking

MorrowCameron:
Final step to spinning up prod. This will add server spin up to our terraform files. It also makes client either manually buildable or auto-build after the server and auth-server to avoid race conditions.

## Vital to squash commits, had to test on pr for terraform

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -48,7 +48,8 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y postgresql-client
 
       # 1.2 Install Supabase CLI
-      - uses: supabase/setup-cli@v1
+      - name: Install SupabaseCLI
+        uses: supabase/setup-cli@v1
         with:
           version: latest
 
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -8,12 +8,14 @@
 # Credit: AlfredMadere
 # Modified: EthanV1920
 
-name: 'Terraform Server Depoly'
+name: 'Terraform Server Deploy'
 
 on:
   push:
-    branches: [ "prod" ]
+    branches: 
+      - main
   pull_request:
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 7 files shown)


.github/workflows/terraform.yml
```diff
@@ -8,12 +8,14 @@
 # Credit: AlfredMadere
 # Modified: EthanV1920
 
-name: 'Terraform Server Depoly'
+name: 'Terraform Server Deploy'
 
 on:
   push:
-    branches: [ "prod" ]
```
MorrowCameron, line 15: We recognized that prod was not actually being pushed to, so autodeploys were based off the wrong branch. We can change this later to add version control.

packages/client/src/services/ablySetup.ts
```diff
@@ -9,7 +9,7 @@ export let playerChannel: Types.RealtimeChannelCallbacks;
 
 const SERVER_URL = process.env.SERVER_URL;
 let channelsBoundToWorld: boolean = false;
-let authorizer = SERVER_URL.slice(-1) == "/" ? 'auth?username=' : '/auth?username='
```
MorrowCameron, line 12: This was causing a linting error. Attack of opportunity to fix it

sql/setup.sql
```diff
@@ -17,7 +17,7 @@ create table
     health smallint not null default 100,
     pname text not null default 'Kyle',
     gold bigint not null default 0,
-    appearance int[] default not null '{16734003, 4367253, 1752748}', -- Random default value for now
+    appearance int[] default '{16734003, 4367253, 1752748}' not null, -- Random default value for now
```
MorrowCameron, line 20: This need to be changed, this change to the setup script is what broke server spin up and terraform for a bit.

MorrowCameron:
If you approve please please please squash and merge so I don't add 20 commits to the repo without needing to. Thank you!

AlfredMadere (approved):
Good work Cameron 🤝
Twas an epic

sql/setup.sql
StarDylan: I agree, not null should line up with the rest of the columns:
```
appearance int[] not null default '{16734003, 4367253, 1752748}',  -- Random default value for now
```

StarDylan (changes requested):
The SQL needs to be fixed before merging @MorrowCameron

sql/setup.sql
MorrowCameron: The not null being before the random default values was causing issues on build and causing the setup.sql file to fail. This was an intentional which was why I pointed it out yesterday. We can try to deep dive this another time but if we don't move the `not null` then the terraform server deploy will fail on creation.

StarDylan: Ok thank you for clarifying this. I'll approve as-is

StarDylan (approved):

Issue #255, opened 2025-01-31
Establish Data Pipeline Between World Servers


IsaacLake03:
The current plans for the portal system, involves hopping into a portal to pick another world to go to. In order to manage this transportation we must first implement a pipeline to transfer player data between servers. This first sub-issue covers creating a data pipeline.

Deliverables:
- Make a function to have the server update its record of a client with its current cached data.
- Make a function to tell the new server to load in the new character (Ensure that the auth server handshake does this)
- Feed the data from the new server to the og client

PR #291, opened 2025-02-06
feat: Establish data pipeline for world_id from client->Supabase

ethanhandelman:
Group: Andrew Estrada, Ethan Handelman, Isaac Lake
Addresses issue #255 

**Feature Summary**
Adds persistence for world_id between client->server->auth-server->supabase. Allows for transition of worlds (through client portals, etc.) to be implemented.

**Changes Made**

Added world id as part of body in REST API call from server->auth-server
`packages/auth-server/src/characterData.ts`
`packages/server/src/services/authMarshalling.ts`

Added world id into message for Ably 'leave' broadcast channel
Added `MAINTAIN_WORLD_OPTION` const, used when player should remain in same world
`packages/client/src/services/playerToServer.ts`
`packages/client/src/services/serverToBroadcast.ts`
`packages/server/src/services/clientCommunication/ablyService.ts`


KyleQ1:
so much documentation is just removed

packages/auth-server/src/characterData.ts
```diff
@@ -3,21 +3,40 @@ import { Request, Response } from 'express';
 
 const characterData = async (req: Request, res: Response) => {
   const id = req.params.Id;
-  console.log('Id:', id);
-  const { health, name, gold } = req.body;
-  if (!id || !health || !name || gold === undefined) {
+  // console.log('Id:', id);
```
StarDylan, line 6: We can probably just remove this line

packages/client/src/services/playerToServer.ts
```diff
@@ -48,9 +48,14 @@ export function startWorld() {
   });
 }
 
-export function leaveWorld() {
-  publishPlayerStateToPersist();
-  broadcastChannel.presence.leave(publicCharacterId, (err) => {
+/*target_world_id should be a text world_id from the worlds table in supabase
+  or 'STAY_AT_WORLD' as defined in serverToBroadcast to indicate staying in the same world */
```
StarDylan, line 52: We should change this to [JSDoc ](https://jsdoc.app/about-getting-started)so it shows up nicely in our IDEs

StarDylan (changes requested):
Some minor style issues, but love the comments that say where to ensure that it matches, it should be very helpful

packages/auth-server/src/characterData.ts
ethanhandelman: Okay, comment deleted

packages/client/src/services/playerToServer.ts
ethanhandelman: Done, good idea

StarDylan (approved):
Looks good!

PR #294, opened 2025-02-06
Persist Server Data

logancostello:
We made it so we server data is stored and persisted in Supabase. The flow looks like the following:
1. When the server starts, it downloads the data files from Supabase. If this fails, then local files are used. These local files are also uploaded to Supabase incase there is nothing there (the logic here should probably be improved upon in the future, but it works).
2. Every 10 minutes the data is uploaded to Supabase
3. When the server receives signal interrupt, it uploads to Supabase before closing

These changes allow players to keep their world. Developers can reset their world by creating their dev world then deleting the files in supabase.

These changes will have instructions for developers to update their .env files and supabase to allow for this to work.


jmdebarro:
Would it be possible to allow for an extension given circumstances, terraform CI?CD issues? Leaving this here for record-keeping.

logancostello:
Our two failing CI / CD errors are both out of our control. 

Terraform Error: Somebody in slack said they caused this error and are working on it.

Tests: The tests fail because the CI/CD pipeline does not already have the new env vars we introduce with this PR. We do not have the ability to add the env var to the CI/CD, but the test runs on our local env with proper variables.

Thanks for understanding

jmdebarro (approved):
Issues could potentially come for local instances if people CTRL+C the handleExit process while it is writing to supabase. This is solved by deleting those malformed files from storage (easy), then rebuilding. Would not be an issue in prod env as a prod server would not forcefully exit it's processes, I believe.

LGTM!!!

logancostello:
> Issues could potentially come for local instances if people CTRL+C the handleExit process while it is writing to supabase. 

I haven't experienced this, but I could imagine if you CTRL+C twice how that could lead to an error. The first should cause an upload, then close, but if they do it again and start another upload and close there could be weird race conditions. Not sure what we can do about that though, considering CTRL+C quite literally quits programs

AndrewJMart (approved):
I propose the same issue as Jonathan. However, I do not see an instance where a server would forcefully close? However, if a server does crash during the upload process that could potentially be a problem???? It could be worthwhile to store a couple of snapshots at a time so not all data is corrupted if something like this does happen (if not for the exact reason listed before potentially another scenario could cause DB corruption). For a local dev environment this is not an issue, however as we migrate to a hosted prod environment this is definitely an issue that would need to be addressed in a future PR. 

In regards to the code itself, the CI/CD pipeline errors are out of our control and the testing does work with the appropriate ENV variables.

LGTM

logancostello:
Updated the code to fix potential issues with force closing the server. This behavior no longer occurs. Instead, developers can use the cheat code SHIFT+S to save the current world state to Supabase. This is a much more elegant solution and is a QOL improvement for developers.

logancostello:
We will update wiki page to include the cheat when we merge

packages/server/src/services/supabaseStorage.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (!process.env.SUPABASE_URL || !process.env.SUPABASE_SERVICE_KEY) {
+    throw new Error('Cannot run without supabase credentials in env.');
+  }
+
+  return createClient(
+    process.env.SUPABASE_URL,
+    process.env.SUPABASE_SERVICE_KEY
+  );
+}
+
```
StarDylan, line 21: The tests should be able to run without the supabase creds in env. Since its an external connection it's probably Ok if we don't test the supabase connection, but the `initializeSupabase()` needs to be put somewhere that only runs when we actually run the server, ideally next to other init logic.

StarDylan (changes requested):

packages/server/src/services/supabaseStorage.ts
logancostello: We successfully moves the initialization to the setup.ts file, but we are still getting issues. We thought of two solutions, but both seem imperfect. Would love to run them by you and see which is preferred.

1. Initialize Supabase when per upload/download. Every time uploadLocalData or downloadData is called we would initialize a new Supabase client. This would isolate the supabase client to those functions. The downside is we would have to reinitialize the client even if we have done it before. This process is fast though.

2. Create a mock supabase instance in the required tests. This would continue to allow the global Supabase instance to exist, but from the test's perspective the import would be shadowed by some mock Supabase instance that does not require the env variables. This would mean that any tests that use files that use the import would have to each individually mock the instance. 

I think the first option is better, but please let us know what you think and if there are other, better options. Thanks

StarDylan (approved):
Looks good

PR #297, opened 2025-02-06
Unified Global JSON

tarannssingh:
1. Created a centralized global.json file that contains all of the information from the sever and client global.json files.
2. Created a script to parse the new centralized global.json and make specific global.json files for the client and server

Taran and Tyler


TylerNP:
Looks good to me

packages/json_splitter/generator.ts
StarDylan, line 1: Maybe we want to integrate this generator script into the build process, so when we run the server it auto-builds the json files (and then we can .gitingore the server and client json files)? What do you think?

With this solution I could potentially see people modifying the wrong script and then their changes being overwritten. As well as when we modify global.json in the generator it also changes global.json for server and client, so those changes are just noise and don't reflect the actual change.

TylerNP: I like this change, but should the generator script run for the once for the server and once for the client. For example, the script would only generate on the server when you run "pnpm build" in the server package and separately the client would only generate when "pnpm" build is called in the client package.

packages/client/static/global.json
StarDylan, line 1: Make sure to remove the artifact with `git rm --cached`

packages/server/data/global.json
StarDylan, line 1: Same for this one

StarDylan (changes requested):
Delete the global.json and fix the main conflict, otherwise looks great!

packages/server/data/global.json
TylerNP: I will first update the unified global json, since some changes have been made.

StarDylan (approved):

PR #299, opened 2025-02-06
New Potions + Bug Fix

eswenke:


eswenke:
Hoping to get a small extension on this so we can spend a bit longer degbugging - Lucas helped us diagnose the problem in lecture yesterday but we spent so much time debugging to little avail + cannot make new potions until this is fixed.

StarDylan (commented):
Sorry, since I don't have any effect on grading I can't offer extensions, you'll need to ask the Professor about that.

eswenke:
outdated code, current PR #360 fixes issues.

PR #304, opened 2025-02-07
Added Custom Domain Name

MorrowCameron:
We want the custom domain `potions.gg` to be used for the server world. Currently, it seems to be getting overwritten on client builds, so we have updated the build script to include the CNAME.

[Docs we followed](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-add-cname-file-cname)


StarDylan (approved):
Looks good

PR #317, opened 2025-02-10
Simplify Upload/Download Data Process & Allow For Multiple Worlds to be Saved

logancostello:
We simplified the upload/download data process by merging the WAL file into the db file, then only uploading the db file. This is not only simpler, but also fixes a rogue bug with the previous implementation.

In addition, the file being uploaded now includes the world ID, allowing for multiple worlds to be saved.


jmdebarro (approved):
lgtm

AndrewJMart (approved):
lgtm

StarDylan (approved):
Looks good

Issue #231, opened 2025-01-30
World Aggro Logic


IsaacLake03:
Current logic ideas are:
Villagers don’t start fights
Fighters only fight blobs (but always fight them)
Blobs fight when hungry (satiation is low) or when you have low favorability with them

Deliverables:
- Ensure that all NPCs follow the logic above

PR #324, opened 2025-02-11
Mob aggro logic

Andrew76xx:
Group: Darren Hong and Andrew Mei

Adding Feature for (Issue https://github.com/sloalchemist/potions/issues/231). 

**Description:** The fighting logic for the game has been implemented in the `utility` function in `hunt.ts`, where Fighters will always fight blobs, blobs always fight when really hungry or have low favorability with the enemy, and villagers do not start fights. Tests were made in pre-existing file `mobPlanningtests.ts`. `flee.ts` has been updated so that the fighter will never run from a fight.


packages/server/src/mobs/plans/hunt.ts
```diff
@@ -37,9 +37,26 @@ export class Hunt implements Plan {
...
+            ((100 - npc.needs.getNeed('satiation')) / 100) *
+              npc.personality.traits[PersonalityTraits.Gluttony] +
+            npc.personality.traits[PersonalityTraits.Aggression] *
+              (npc.attack / this.enemy.attack);
+        }
+      }
+    }
+    if (npc.type == 'fighter') {
+      utility = 100;
+    }
```
StarDylan, line 59: We ideally don't want to add mob-specific logic into these general classes. What if we create a new mob type that also should never start fights? These should ideally go in the global.json somehow.

packages/server/src/mobs/plans/flee.ts
```diff
@@ -54,7 +54,7 @@ export class Flee implements Plan {
   }
 
   utility(npc: Mob): number {
-    if (!npc.position) return -Infinity;
+    if (!npc.position || npc.type == 'fighter') return -Infinity;
```
StarDylan, line 57: Same feedback here

StarDylan (changes requested):

packages/server/src/mobs/plans/hunt.ts
DarrenH6293: So would we add categories like `"passive_mobs": ["villager"]` to the global.json and in the plans, check for `passive_mobs`?

Andrew76xx: @StarDylan

StarDylan: That looks good, sorry for not responding, I probably need to adjust my notifications so I actually get them.

StarDylan (approved):
Looks good!

PR #328, opened 2025-02-11
fixed: container capacity

### LINKED ISSUE #14: Make Containers Limit Capacity
Baskets have infinite storage, despite a set capacity in the code.

Deliverable:
Change to container.ts to prevent adding of items over capacity.

Meboy1000:
Fixed #14 

Added a proper check to the container.ts file to limit capacity to its stated value

Had a ton of issues with merge conflicts for no good reason so I just force reset it to main and re added the changes since they were super small, unfortunately also lost all the history from that but it just wasn't working and I am the only one really affected by it so.

### DIFF
--- a/packages/server/src/items/container.ts
+++ b/packages/server/src/items/container.ts
@@ -26,7 +26,7 @@ export class Container {
   }
 
   getCapacity(): number {
-    return 10;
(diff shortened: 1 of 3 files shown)


StarDylan (approved):
Looks great! Good catch

PR #334, opened 2025-02-12
Update Node Dependency + Make Supabase Process with Terrafom

logancostello:
- Updated the node dependencies to a newer version where File is supported.
-  Removed the need for developers to manually create a bucket on Supabase. Instead code automatically generates it
- Added previously added environment variables to terraform


jmdebarro (approved):
This is better, nice refactor. lgtm, this time surely.

AndrewJMart (approved):
Supabase code looks senior dev level… 

terraform implementations will suffice.

Nice programmatic implementation and should integrate cleanly with current people’s current setup

StarDylan (approved):
Looks good!

PR #338, opened 2025-02-12
Feature: Added Interactions for Water World

spruhaN:
- Seaweed drops seaweed pieces that can build walls
- Volcano drops fireballs that can be consumed or made into potions
- Clams drop pearls that can be made into potions
- Message in a Bottle prompts to Read Message (needs further conversation before fully implementing)
- Added graphics to support this


packages/client/src/scenes/worldScene.ts
```diff
@@ -105,7 +105,17 @@ export class WorldScene extends Phaser.Scene {
...
+    //     key: `volacano`,
+    //     frames: this.anims.generateFrameNames('global_atlas', {
+    //       start: 1,
+    //       end: 8,
+    //       prefix: `volcano-`
+    //       //suffix: '.png'
+    //     }),
+    //     frameRate: 6,
+    //     repeat: -1
+    //   });
```
StarDylan, line 118: We should just delete this code instead of commenting it out (it adds noise to the code base)

StarDylan (changes requested):
Small style thing, otherwise looks good

StarDylan (approved):

PR #339, opened 2025-02-12
Implement "Enter Portal" interaction and allow user to select a target world

AdamMikacich:
**Team**
Adam Mikacich, Brett Hillyard, Kannan Jain, Kyle Quinlan

**What changed**
We implemented the "Enter Portal" interaction that displays all available worlds that the user can connect to. The portal now appears as open. When the user interacts with the portal, the server validates they are actually next to a portal (to prevent cheats), and then responds with a list of available worlds. The list of worlds is fetched from the `auth-server` package by the `server` package. The event is then added to the broadcast to trigger the scene for the user. A new overlay scene renders the options and allows the user to select the world they intend to join. Upon clicking a world, the desired world id + name is logged to the console to allow the other group to continue their world-transition implementation.

**Why**
The user needs to be able to view and select a world to transition to after interacting with the portal.

**Screenshots**
Open portal appearance:
Screenshot 2025-02-12 at 12 55 37 AM

Enter portal interaction:
Screenshot 2025-02-12 at 12 55 43 AM

Portal list:
Screenshot 2025-02-12 at 1 04 36 AM


StarDylan (approved):
Looks great!

PR #353, opened 2025-02-12
fix: prevent give item interaction if mob is carrying

ruiyichi:

### Bug
When a player possesses an item and encounters a mob that already holds an item, the option to give the item to the mob should not appear. Functionally, nothing happens if the button is pressed.

### Changes
https://github.com/sloalchemist/potions/commit/4d9c9ffddc882e8206f310730cab5b9c793abf58

- Check if mobs are carrying items before adding the 'give' interaction

### Testing

- Created `playerCarriedItems.test.ts` to verify presence/absence of 'give' interaction


StarDylan (approved):
Looks good!

PR #359, opened 2025-02-13
fix max health being attack value after death

KeithJLZ:
After dying for the first time, player max health becomes the value of their attack.

Fix: Set player health to the correct value in `packages/server/src/services/clientCommunication/ablyService.ts` when sending player data after dying instead of setting it to the player's attack value.

Unit Tests: None


StarDylan (approved):
Looks good

PR #368, opened 2025-02-13
Terraform dev improvments (amended)

AlfredMadere:
- Removed need for retry-script
- Automatically applies supabase migrations
- Dockerized dev terraform 
   -  Devs are no longer required to install anything except Docker to run terraform
   - Terraform now runs exactly the same for everyone regardless of machine 
- Now catching known errors (like incorrect permissions on key) and giving devs tips on how to solve in the error itself
- Updated readme to match


NickPerlich (approved):
After following the instructions to donwload docker desktop, tf.sh and tf.ps1 both work to spin up and create resources as a replacement for terraform apply.

AlfredMadere:
Had 5 different people in the class try out these changes and everything went smoothly

StarDylan (approved):
Looks good!

PR #379, opened 2025-02-13
Logging ticks, number of mobs, number of items into SQLite db

kailaaquino:
### Feature:
Created "grafanadata" table in `packages/server/data/server-data.db` that tracks the tick_id, number of items, and number of mobs on the local game state. 

We created the table and made insertions in `packages/server/src/grafana/dataLogger.ts`
This table will be used for Grafana visualizations using the SQLite plugin. Later version may move to Prometheus.


StarDylan (approved):
Love logging for Grafana! Can't wait to see the visualizations!

PR #384, opened 2025-02-13
Quick Fix for Global.json

irwinj22:
added tree and log back to global.json (in both server/ and client/). will develop more comprehensive solution soon.


jackalnom (approved):
ship it!

Issue #233, opened 2025-01-30
Decrease favorability on fighting


IsaacLake03:
When combat is initiated, targets favorability of you decreases by 20 points

Deliverables:
- Decrease species' favorability upon attacking one of their NPCs

PR #386, opened 2025-02-13
233 decrease favorability on fighting

TheAllenChoi:
Closes: #233 
- Implements a system where favorability decreases upon hitting a mob
- Updates favorability upon hitting a mob
- Test cases for all of the above


StarDylan (approved):
Looks good

PR #388, opened 2025-02-14
262 Fix bug 313 dropped items can't be picked up + test, + player join diagram

ethanhandelman:
Group: Ethan Handelman, Andrew Estrada, Isaac Lake

### Bug Description
When a player/mob is carrying an item, and another item is available to pickup, if the player does so, the original item that is dropped is unable to be picked up again. 
### Changes Made
Added logic to check for if mob is currently carrying item, and drop if so 
`packages/server/src/items/uses/pickup.ts`

Added test for this situation
`packages/server/test/items/itemPickupWhileCarrying.test.ts`

Removed unnecessary log on item pickup
`packages/client/src/world/item.ts`

### Documentation Added
Added data persistence diagram describing flow of interactions between the server, auth-server, client, and Supabase that allow a player to join the server with attributes set by data in Supabase. Added a more detailed description in GitHub wiki:
[]http://Player%20Data%20Persistence%20for%20Player%20Join


packages/server/src/items/uses/pickup.ts
```diff
@@ -25,6 +25,10 @@ export class Pickup implements Use {
       return true;
     }
 
+    if (mob.carrying) {
+      //drop item if currently carrying one
+      Carryable.fromItem(mob.carrying)!.dropAtFeet(mob);
+    }
```
StarDylan, line 31: Note the code coverage notice. It seems your test isn't actually running your change. I.e. if we remove this code, then the test still passes.

So the test might not be testing the actual bug, or it might not be fixed (or the bug was already fixed and this code isn't doing anything)

StarDylan (changes requested):

packages/server/src/items/uses/pickup.ts
IsaacLake03: Fixed the test issue, now everything is covered

StarDylan (approved):
Looks good

Issue #364, opened 2025-02-13
Effects UI stays when switching to the Fight tab


DarrenH6293:
![Image](https://github.com/user-attachments/assets/ddba047a-66ff-4b4d-a932-f4c23d95931b)

The effects UI stays when switching to the Fight tab.

Deliverables:

- Make sure the effects UI disappears when switching to all tabs.

PR #391, opened 2025-02-14
fix: The effects UI is set to not visible when switching to fight tab

DarrenH6293:

Bug Fix for #364 

**Description**: The effects UI for the Handbook tab stayed when switching over to the Fight tab. We fixed this by setting the problematic containers to not be visible when switching to the Fight tab in `uxScene.ts`.

Before:
image

After:
image


StarDylan (approved):
Looks good!

Issue #394, opened 2025-02-14
Client-side Permissable Interactions Refactor


chelseylearns:
Previously, the controller was using global variables to determine which interactions should be available to players on the client side. 

This refactor makes it such that ownership identifiers are passed into getPhysicalInteractions(), making it simpler to implement individual ownership. Also this will make unit testing cleaner.

PR #397, opened 2025-02-14
Refactor #394: Client-side Permissable Interactions

chelseylearns:
## Summary
Previously, the controller was using global variables to determine which interactions should be available to players on the client side.

This refactor makes it such that ownership identifiers are passed into getPhysicalInteractions(), making it simpler to implement individual ownership. Also this will make unit testing cleaner.

## Changes
- modified `packages/client/src/world/controller.ts`: Refactor getPhysicalInteractions to recieve an ownerId to determine permissible interactions such that it is not relying on global variables.
- modified `packages/client/src/sprite/sprite_mob.ts`: Add id field to sprite_mob.
- modified `packages/client/test/items/uses/playerPermissibleInteractions.test.ts`: Remove globally defined mock variables.
- modified `packages/client/test/items/uses/smashGate.test.ts`: Revert to original stats.

### DIFF
--- a/packages/client/src/sprite/sprite_mob.ts
+++ b/packages/client/src/sprite/sprite_mob.ts
@@ -29,6 +29,7 @@ export class SpriteMob extends Mob {
   speechBubbleHeight?: number;
   bubbleOffsetX?: number;
   bubbleOffsetY?: number;
+  id?: string;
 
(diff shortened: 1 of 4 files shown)


StarDylan (approved):
Looks good!

Issue #363, opened 2025-02-13
Can only interact with 1 basket at a time


DarrenH6293:
![Image](https://github.com/user-attachments/assets/e6ab55f1-6a87-490f-920e-e3de72a9abec)

Can only interact with one basket at a time, if there are multiple nearby baskets near the player. The player has to move and make only 1 basket near them to ensure the correct basket interact option shows up. This bug causes inconvenience.

Deliverables:

- Make sure the player have the options to interact with all baskets near them.

PR #400, opened 2025-02-14
Basket interactions bug fix

DarrenH6293:
**Group**: Darren Hong, Andrew Mei

**Bug Fix: #363**

**Description**: Modified the `getInteractablePhysicals` function in `controller.ts` so that the items tab showed all the nearby baskets as interactable, instead of just the nearest one.

Before:
image

After:
image


StarDylan (approved):
Looks good!

PR #401, opened 2025-02-14
Fix: Implement server-side authorization for server to auth-server communication.

AdamMikacich:
**Names:** Adam Mikacich, Brett Hillyard, Alfred Madere

**What changed:**
Before this fix, both the `characterData` and `worlds` endpoints lacked any authorization checks. This means that anyone with the endpoint path could successfully process a request. We resolved this security vulnerability by creating a shared secret in terraform that only the `server` and `auth-server` packages know about. The `auth-server` now validates that all requests have this shared secret injected into the Authorization header as a Bearer token.

**Why:**
The `characterData` and `worlds` endpoints were publicly accessible, meaning that anyone could manually update character data (such as their gold amount) or read all accessible worlds.


StarDylan (approved):
Awesome looks great!

PR #402, opened 2025-02-14
Fix: Validate environment variables at runtime to prevent silent failing

AdamMikacich:
**Names:** Adam Mikacich, Brett Hillyard, Alfred Madere

**What changed:**
When using environment variables, we now access them from the `getEnv` function inside the new utils.ts file in the common package. This helper validates that the value exists before usage, otherwise an error is thrown. This allows the application to "fail early" to more easily find any issues. Additionally, this has the benefit of typing the environment variable as `string` (instead of `string | undefined`) without casting due to the runtime check.

**Why:**
Previously, when new environment variables were created and used, the build could silently fail if they were not properly set. This change checks for all environment variables when referenced at runtime to avoid silent failure.


StarDylan (approved):
Looks good

Issue #244, opened 2025-01-30
Market Favorability Logic


idavoong:
If a player has high favorability with certain species, species are more inclined to trade/buy from the player. However, if the price is too high, this effect doesn't apply.

Deliverables:
- Implement a system for the likelihood that an NPC will buy from a player's stand based on affiliation

PR #411, opened 2025-02-16
244 - Market Favorability Logic Added

sreshtalluri:
## Feature

If a player has high favorability with certain species, species are more inclined to trade/buy from the player. However, if the price is too high, this effect doesn't apply.

## Implementation

Implemented a favorability-based pricing system to make purchases more dynamic and responsive to relationships between species. At a neutral favor level of 50, the max price remains 20, but now, for every point below 50, the max price decreases by 0.2, making unfriendly species more price-sensitive. Conversely, for every point above 50, the max price increases by 0.5, allowing highly favorable species to tolerate higher prices. I also added a minimum price cap to prevent species from being overly stingy and a maximum cap to ensure that even highly favorable species won’t overpay. 

Additionally, I introduced a hard price rejection rule—if the item's price exceeds the max cap, favorability and affiliation no longer matter, and the purchase is automatically denied. These changes create a more balanced and realistic trading system while preventing exploits.

These values and thresholds can be changed/adjusted later on depending on the ranges of favorability.

## Tests

Added 5 tests for different scenarios:

1. Low favorability, reasonable price -> should not buy
2. Neutral favorability, reasonable price -> should buy
3. High favorability, slightly higher price -> should buy
4. High favorability, extra high price -> should not buy
5. High price -> should not buy

Issue [#244](https://github.com/sloalchemist/potions/issues/244)


StarDylan (approved):
Looks great!

PR #414, opened 2025-02-17
feat: add tests for SpriteHouse functionality and mock dependencies

Fenlongo:


packages/json-generator/dist/generator.js
StarDylan, line 1: I think this build artifact can be removed. (with `git rm --cached <filename>`)

packages/converse/package.json
StarDylan, line 1: Also could you explain why this was changed, as it seems tangential to your PR.

StarDylan (changes requested):

Issue #210, opened 2025-01-29
Add Expanded Inventory


anvch:
The player should be able to pick up multiple things at a time and use them. Currently, the inventory should be 5 slots, unstackable (if you can make it easy to extend later on, please do so because this is not super set in stone).

Deliverables:
* Inventory panel (5 slots, unstackable) in-game.
* Interactions using the inventory:
     * On picking up an item, it should automatically be added to the inventory.
     * Have some way to do actions with a specific item in the inventory (give it away, eat, drop, make it a potion, etc.) -- select it, then perform the action or something similar.
* Have inventory information for a player stored in the database.

PR #415, opened 2025-02-17
Add expanded inventory. Issue #210

apervmm:
We have added the logic of adding Item into the inventory and being able to hold multiple objects simultaneously. Player now able to stash the object into an inventory once the object was picked up and then drop the selected item from inventory. 

Screenshot 2025-02-17 at 3 01 19 PM


packages/client/dist/bundle.js.LICENSE.txt
```diff
@@ -30,14 +30,14 @@ PERFORMANCE OF THIS SOFTWARE.
...
+ * @copyright    2013-2025 Phaser Studio Inc.
  * @license      {@link https://opensource.org/licenses/MIT|MIT License}
  */
 
 /**
  * @author       Florian Vazelle
  * @author       Geoffrey Glaive
- * @copyright    2013-2024 Phaser Studio Inc.
+ * @copyright    2013-2025 Phaser Studio Inc.
  * @license      {@link https://opensource.org/licenses/MIT|MIT License}
```
StarDylan, line 41: We should probably remove this bundle.js.LICENSE.txt from the repo using `git rm --cached <filename>`, also add it to the gitignore so it doesn't get tracked again in the future.

packages/client/src/services/serverToBroadcast.ts
```diff
@@ -196,8 +215,22 @@ export function setupBroadcast(
           handleGiveItem(broadcastItem.data as GiveItemData);
           break;
         case 'drop_item':
+          // console.log(broadcastItem.data as DropItemData, "DROP")
```
StarDylan, line 218: Also remove this comment, as its just noise in the codebase

StarDylan (changes requested):
Just some style fixes

packages/client/dist/bundle.js.LICENSE.txt
apervmm: image

apervmm: image

apervmm: So now, it's not even different from origin/main.

packages/client/src/services/serverToBroadcast.ts
apervmm: Same with this, everyones commented parts gets pushed to codebase, which means they were not bounced, but our did because of it.

packages/client/dist/bundle.js.LICENSE.txt
apervmm: Yes, I agree that Build files should never be in codebase and needs to be gitignored, but I think it might be unfair to bounce exactly our PR only because of the Build file, while you have approved for all other groups

StarDylan (approved):

PR #371, opened 2025-02-13
Feature: New Items Added To Global (part 1)

sribala20:
First edition of new items!
* New sprites for items listed on items issue
* Players can interact with and brew items
* Spawning is controlled by static world 
** Portal spawns Bomb, Gun Powder, and Sands of Time
** Heartbeet-plant: bones, slime blob
** Blueberry: tar, lightning bloom
** Tree/seaweed: Sunflower, Sun Drop 

To be continued:
* Updating interactions (between players and mobs)
* Improved spawning

PR #430, opened 2025-02-19
Improvement: Village World Spawning Improved

sribala20:
* New spawners added in village world to fix current spawning strategies
* Bones and Slimeblobs spawn on deaths (1/10 times)


irwinj22 (approved):
lgtm 💯 🛵

packages/client/static/global.json
```diff
@@ -59,13 +59,7 @@
       ],
       "walkable": true,
       "smashable": true,
-      "interactions": [
-        {
-          "description": "Eat",
-          "action": "eat",
-          "while_carried": true
-        }
-      ],
```
StarDylan, line 68: Also please comment on why this was removed, as it seemingly doesn't have to do with spawning.

packages/server/src/mobs/mob.ts
```diff
@@ -612,14 +612,32 @@ export class Mob {
         // is currently incomplete on mainline) will update the changeGold
         // method to persist to supabase in addition to the local DB
         this.changeGold(-halfGold);
+      }
+      if (this.type === 'blob' && Math.random() < 0.1) {
+        itemGenerator.createItem({
+          type: 'slime-blob',
+          position
+        });
```
StarDylan, line 620: Hard Coding this into mob.destroy is probably not the best way to do this. Since if other mobs want to also have similar functionality now we have to add lots of if statements, maybe add a loot table (See [Minecraft Loot Table](https://minecraft.wiki/w/Loot_table)  for inspiration) to the config?

packages/server/src/mobs/mob.ts
```diff
@@ -612,14 +612,32 @@ export class Mob {
...
+            position
+          });
+        }
       }
+      // this.pendingPositions.push(position);
+      // console.log(this.pendingPositions);
+      //   if (this.pendingPositions.length >= 4) {
+      //     this.smashItemsAtPositions(this.pendingPositions, this);
+      //     this.pendingPositions = [];
+      //   }
```
StarDylan, line 640: This can also be removed, as it adds noise to the codebase

.github/workflows/terraform.yml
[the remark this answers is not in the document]
irwinj22: oh that was an accident. thanks for catching!

packages/client/static/global.json
irwinj22: this was removed as clean-up from the [part one of our PR](https://github.com/sloalchemist/potions/pull/371). bones shouldn't have been edible in the first place, so we thought it was best to correct that now.

packages/server/src/mobs/mob.ts
sribala20: Thanks for pointing this out Dylan! We decided to remove that functionality-- especially since slime blobs were meant to spawn in oozon/ water world anyways. We reconfigured bones to spawn from the cannons now instead. Hope this works.

melissachow0 (approved):
lgtm!!

irwinj22 (approved):
🛵 💨

StarDylan (approved):
Looks great! Thanks for your comments!

Issue #431, opened 2025-02-19
Implement mob favorite item


TheAllenChoi:
- Mobs should have a favorite item assigned to them upon creation
- Favorite item should be random, not dependent on species
- Favorite item should be able to be accessed by other functions

PR #436, opened 2025-02-19
Implement favorite item for Mob

TheAllenChoi:
- Mobs should have a favorite item assigned to them upon creation
- Favorite item should be random, not dependent on species
- Favorite item should be able to be accessed outside

Closes: #431


StarDylan (approved):
Great PR!

PR #440, opened 2025-02-19
Disable movement and tab-menu clicks when Enter Portal Menu is open

### LINKED ISSUE #437: Disable world clicks when "Enter portal" menu open
When the "Enter portal" menu is open, players can still click to move on the world. Instead, we want them to only be able to click the available buttons on the menu.

xgarcia17:
**Developers: Xavier Garcia, Dre Seguitan**

Closes #437 

*Changes:* When the "Enter portal" menu is open, we don't want users to be able to click things below/underneath/behind the menu. So, we added a "blocking" layer to the game when the "Enter portal" menu is opened. This layer consumes all clicks that would otherwise go to the layers below it (i.g. the ground to set targets for movement, the tabs (i.g. Info, Items, etc.)).

*Testing:* This is a visual bug, so we tested our code by simply running the game and trying to click things that shouldn't be clickable (movement and tabs) and ensuring that the "close" button (which is currently the only button on the menu so far) is still clickable and closes the menu. Additionally, once the menu is closed, other clicks once again work as intended.

*Files Changed:* packages\client\src\scenes\portalMenuScene.ts in create()


StarDylan (approved):
Looks great!

Issue #366, opened 2025-02-13
Item Ownership Not Persisting For Potion Stand


chelseylearns:
Immediately after a potion stand is constructed, the item ownership is current and all player options are available for interaction. 

Shortly after, for unknown reasons, item ownership appears to be undefined and interactions are unavailable. We need to investigate the cause of this (potentially villager activity nearby) and solve before ownership can be resolved. 

Deliverables:
- Make item ownership persist for as long a player's potion stand is standing

PR #441, opened 2025-02-19
Fix Bug #366: Item Ownership Not Persisting for Potion Stand

chelseylearns:
## Summary:

This fixes the following bug: #366 

Immediately after a potion stand is constructed, the item ownership is current and all player options are available for interaction.

Shortly after, for unknown reasons, item ownership appeared to be undefined and interactions were unavailable. We found the cause of this and implemented a fix. 

## Changes
- modified `packages/server/src/services/clientCommunication/clientMarshalling.ts`: added items_owned_by field to getItemsAbly sql query so that item ownership will persist


StarDylan (approved):
Looks good!

PR #450, opened 2025-02-19
fix: reset movement key state after move

ruiyichi:

### Bug
When tapping the WASD keys to move, the player sometimes moves two tiles instead of one. This occurs because the keypress state remains true over multiple ticks.

### Changes
https://github.com/sloalchemist/potions/commit/ccd7924ea5e18173eda8042d95cba7e19795a98b
- Reset the state of the movement keys after the player moves


StarDylan (approved):
Looks good!

PR #452, opened 2025-02-19
Bug Fix: Update Vscode Task Script

spruhaN:
With the new terraform changes the output on the auth server is different which then does not trigger the script properly.
- Changed the string that triggers the end of the auth-server monitoring and starts start the server and client stages.


StarDylan (approved):
Looks good! Nice and clear description

PR #453, opened 2025-02-19
Feature: Merged Client assets and atlas

spruhaN:
To support the concept of transferring world-specific items across different worlds, assets must be accessible regardless of the current world. We addressed this by combining all assets into a single file while eliminating redundancies.

Key improvements:
	1.	Merged the three global.png files from the Villager, Fire, and Water worlds.
	2.	Removed duplicate assets across the PNGs (e.g., portal, sprites, etc.).
	3.	Created a representative atlas for the global PNG. (some items need renaming but in contact with world representatives to fix in a future PR)
	4.     Added files in the world_assets folder so accessible via Github Pages

Additional work has been completed in the _merge_assets_ branch, but further refinement is needed, along with updates to the command-line pull request.


StarDylan (approved):
Looks good

PR #462, opened 2025-02-20
Add More Potion Effects from Potions

anvch:
Added gold potion and grey potion effect as per the wiki.

Gold potion - can increase max health up to 5 times (20 points for each consumption)
Grey potion - the first enemy hit after consumption will have their speed decreased by 50% for a minute


StarDylan (approved):
Looks good!

PR #469, opened 2025-02-20
undefined value fixed

eswenke:
undefined defense value in the UX has been addressed and shows up properly now.


StarDylan (approved):
Looks good!

PR #471, opened 2025-02-21
Fix: added permissions to baskets 

gabrielriedel:
Added permissions to baskets in respective global.json files to prevent non-affiliated species from having the button to add to baskets. I wrote unit tests, but here is UI/UX proof that the buttons correctly disappear now.

## Before:

Screenshot 2025-02-20 at 10 01 15 PM

## After:

Screenshot 2025-02-21 at 12 17 29 AM


StarDylan (approved):
Looks great! Thanks for including the pictures!

PR #477, opened 2025-02-22
Fix: 'get' button appears only if affiliated

gabrielriedel:
Added permissions to the get_item interaction on baskets to only allow affiliated players to 'get.' This removes the get button from appearing to un-affiliated players.

## Before

Screenshot 2025-02-21 at 1 05 07 PM

## After

Screenshot 2025-02-21 at 1 21 37 PM


packages/json-generator/dist/generator.js
StarDylan, line 1: This should be removed, as its a build artifact. (`git rm --cached <file>`)

StarDylan: Probably also check that it is in the gitignore so it isn't re-added

StarDylan (changes requested):
Looks good, just some accidental build files being added

StarDylan (approved):
Looks great! Thanks for adding removing the build files

PR #481, opened 2025-02-22
Fix: Adding Items to Basket Checks the Types Properly

oprylord:
**Summary**
The type check between a basket's assigned item type and the player's carried item type was preventing the adding of items into baskets. 

**Changes**
Fixed the type check to properly compare the right attributes.


oprylord:
The test situation is kind of funny:
The test checks for an alchemist player's inability to interact with a silverclaw basket (obviously with the intent that community checking would be in place).
Up until this PR, an alchemist player **couldn't** interact with a silverclaw basket, but only because **they couldn't interact with baskets at all** due to the item type check never properly comparing and therefore denying use of the basket.

With this PR, players can interact with baskets, but because community checking isn't in place yet, it fails the test for now.
#477 implements the community checking needed for the test to pass.

StarDylan (changes requested):
> The test situation is kind of funny: The test checks for an alchemist player's inability to interact with a silverclaw basket (obviously with the intent that community checking would be in place). Up until this PR, an alchemist player **couldn't** interact with a silverclaw basket, but only because **they couldn't interact with baskets at all** due to the item type check never properly comparing and therefore denying use of the basket.
> 
> With this PR, players can interact with baskets, but because community checking isn't in place yet, it fails the test for now. #477 implements the community checking needed for the test to pass.

Probably the best course of action is to comment out that test, adding a comment that it will be re-added with PR #477, since we can't merge a PR that will break main.

StarDylan (approved):

PR #486, opened 2025-02-23
Fixes water world and fireworld stack error - global.json 

### LINKED ISSUE #484: Fix: water-world and fire-world loading errors
- running the commands to create water world results in a stack error
- investigating errors with the json files now that links aren't hardcoded

Image

halliec6:
Closes #484 

- can now build water world and fire world (currently neither world loads, undefined item type error)
- uses the same global json since there is one unified global.json for all worlds

- we have discussed future considerations for using one global.json vs having each world have its own in world-assets, however for the time being this gets both worlds working on main again


0felicia0 (approved):

Parshana007 (approved):

StarDylan (approved):
Looks good!

PR #503, opened 2025-02-24
fix: update character animations after color change

ruiyichi:

### Bug
When changing either the eye, belly, or fur color from the Customization tab, the character's color doesn't update immediately. 

### Changes
https://github.com/sloalchemist/potions/commit/7875937e128e2a01eff246416173cf0b4def8db2

- Update the character animations after changing the character's color


StarDylan (approved):
Looks good!

PR #508, opened 2025-02-25
feat: added enter portal interaction to jsonGenerator

jmdebarro:
jsonGenerator was missing portal interaction.

This caused all global.json's generated to not contain this.

In turn, causing the interaction to not exist for the portal.


StarDylan (approved):
Thanks for fixing this

Issue #496, opened 2025-02-24
Bug: Sun Located within Villager Area


irwinj22:
Currently, the sun (which spawns sun-drops) is located within the village, which is only accessible to players with "villager" status. The sun should be moved so that sun-drops are accessible to all players.

Issue #494, opened 2025-02-24
Bug: Unable to Pass Through Open Gate


irwinj22:
Currently unable to pass through open gate into villager area.

jqueja:
That's not a bug. If you were to undo this, the client and server will not match up. The server only accepts villagers who are of the type villager not players at this moment.

Not part of the village bro, gotta get that village status - Pierce

Please refer to this [PR.](https://github.com/sloalchemist/potions/pull/425)

irwinj22:
oh i see, thank you. so you have to be a "villager" to get into the village?

jqueja:
yes you have to be part of the villager to get into that village unless you smash the fence.

irwinj22:
makes sense. i'll close the issue then, thanks again.

PR #509, opened 2025-02-25
fix: move sun outside of villager world

irwinj22:
As explained in issue #[496](https://github.com/sloalchemist/potions/issues/496), the sun (which spawns sun-drops) was placed inside the village, which is only immediately accessible to players who are "villagers" (as [explained to me by Josh Queja](https://github.com/sloalchemist/potions/issues/494). Though it would be possible for non-villagers to smash gate/fence to access village, we felt that this was burdensome and not fun playing.


sribala20 (approved):
amazing addition

PR #516, opened 2025-02-25
Add Default Effects to Undefined Potions

### LINKED ISSUE #442: Add default effects to undefined potions
Players are able to create any number of potions with all possible ingredients, including potions that are not defined in the recipe book. We want to make sure that every potion has some effect.
Each undefined potion can fall into three possible categories:
* Potion color is similar to a known potion recipe. Will produce a muted version of the similar potion effect.
* Potion color is black, meaning all r, g, and b values are individually less than 11%. Potion will greatly reduce players health
* Everything else. Will be given a random effect.

Deliverables:
* Add default effects to the undefined potions based on their categorization

charm-13:
## Description
Players are able to create any number of potions with all possible ingredients, including potions that are not defined in the recipe book. We want to make sure that every potion has some effect.
Each undefined potion can fall into three possible categories:
**Potion color is similar to a known potion recipe.** They can be given:
* 50% of the effect if the color is very similar, or
* 30% of the effect if the color is mostly similar

**Potion color is black**, meaning all r, g, and b values are individually less than 28. This potion will reduce the player's health to zero, thus killing them.

**Everything else** will be given a random effect. Random potion effects include:
* 75% speed for 20 ticks
* Health reduced by 20 (or to 1)
* 75% defense for 60 ticks
* 75% attack for 60 ticks
* Max health reduced by 5 (or to 1)
* Speed, attack, and defense get a 15% boost for 60 ticks
* Player is stunned (20% speed) for 10 ticks
* Player loses 5 gold, or all if they have less than 5 gold.

## Problem
Closes #442 

## Impact
All wiki documentation has been updated [(Potion Brewing)](https://github.com/sloalchemist/potions/wiki/Potion-Brewing)

## Testing
Tested effects in game.
Added new unit tests in server tests. Tests include tests
They can be found in `/potions/potionEffects/`. Specific test files include `randomEffects.test.ts`.


StarDylan (approved):
Looks great! There are some cases not covered by the tests, (the deterministic ones, the random ones I think I'm okay with not testing), but that can be covered in a future PR since the logic is pretty clear to me.

PR #519, opened 2025-02-26
Bug: Inconsistent Movement 

jqueja:
## Description
The player’s movement behavior has been fixed so that when any of the movement keys ('W', 'A', 'S', 'D') are pressed, the game will check for obstacles in the chosen direction. If there’s an obstruction, it will not move and the players position will be stagnant.

## Problem
Given the nature of a blocked path, the player would around around the object. Not the intended behavior given the movement keys.

## Context
If the path is blocked, the player's movement will remain stagnant, and the `getItemAt` function from `client/src/world/world.ts` is used to determine if there is something in the way, checking for obstacles or blocked areas. The built-in `isWalkable` function from `client/src/world/item.ts` helps ensure that the player can only move through walkable spaces.

## Impact
Changes the movement for players to be more reliable.

## Testing
Unit Test not needed. Tested the functionality by placing a wall in the middle of the map and made sure that when a key was pressed it did not walk around that object.

### DIFF
--- a/packages/client/src/scenes/worldScene.ts
+++ b/packages/client/src/scenes/worldScene.ts
@@ -620,6 +620,10 @@ export class WorldScene extends Phaser.Scene {
 
     const target = { x: roundedX, y: roundedY };
 
+    // NOTE: the code in the 'else' block moves the player on the client side
+    //       publishPlayerPosition() calls that code itself, so player will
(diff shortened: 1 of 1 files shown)


StarDylan (approved):
Looks good

Issue #507, opened 2025-02-25
Bug: Stashing Items overflows after 15+ items


sribala20:
Currently, there is no maximum to the number of items that players are able to stash. Thus, it is possible to stash more than 15 items and create buttons that populate off-screen: 

<img width="530" alt="Image" src="https://github.com/user-attachments/assets/e4a17d87-91d5-48f5-877c-b6ea791801e6" />

Moreover, we feel that, logically, it makes sense to cap the number of items that a user can store. Thus, we are going to implement that maximum to fix issues in UI and game logic.

PR #520, opened 2025-02-26
507 bug stashing items overflows after 15 items

irwinj22:
## Description
Previously, the stash UI would overflow with buttons when more than 15 items were added. Now, a threshold of 12 items has been set, so the user is not able to generate buttons that do not fit within the screen. Additionally, a counter has been added to the stash UI so that the user always knows exactly how many items are in their stash. 

## Problem
Closes [#507](https://github.com/sloalchemist/potions/issues/507)

## Context
One design decision worth considering is the cap of 12 items. Why 12? Well, we felt that it made sense to cap the number of items at some number (stashing infinite items does not feel reasonable) and 12 buttons are able to populate nicely within the UI. Any more than 12 and we would have run out of space. 

## Impact
Impact is minimal. Though this change does affect a feature that was developed by @apervmm and @joshualeeee, they were contacted privately and approved our work.

## Testing
Testing was verified manually and codebase still passes tests in server/test/items/stashAndUnstash.test.ts

Before change: 

Screenshot 2025-02-24 at 8 49 54 PM

After change: 

Screenshot 2025-02-25 at 11 20 04 PM
Screenshot 2025-02-25 at 11 38 29 PM


sribala20 (approved):
impressive work

StarDylan (approved):
Another potential discussion is to include the logic on the server so it can also validate that no one has more than 12 items (maybe also move the max stash to the config). But we can add another issue to do that if it makes sense.

Otherwise it looks good, thanks for the discussion on the max stash cap, helps to add some context into the PRs.

PR #522, opened 2025-02-26
Tick Debugging Logic

### LINKED ISSUE #521: Feat: Add tick debugging and logs
Add tick debugging and logs to pinpoint bottlenecks in tick logic. 

Allow for options to toggle console or file output on or off.

smbock42:
## Description

Ticks are running slow and we don't know why. This PR will investigate, debug, and pinpoint slow points in ticks to optimize and make it run faster. 

## Problem

Closes #521 - we don't know performance of ticks currently.

## Context

Creating new debugging logic - added options to turn on or off outputs and time measurements to minimize distractions. 

## Impact

 

Unconfirmed whether it breaks/adds a new bug about "Cannot have conversation with ‘self’"

image

## Testing

Only added debugging and time calculation - no tests needed. 

image


smbock42:
After a minute of idle running with the debugger, mob ticks are taking over 100 ms to complete a tick. 

image

Creating new sub issue of #521 to address this...

packages/server/src/services/gameWorld/serverWorld.ts
```diff
@@ -12,6 +14,40 @@ import { UsesRegistry } from '../../items/uses/usesRegistry';
 import { OnTickRegistry } from '../../items/on_ticks/onTickRegistry';
 import { DB } from '../database';
 import { DataLogger } from '../../grafana/dataLogger';
+import fs from 'fs';
+import path from 'path';
+
+const DEBUG_TO_CONSOLE = false;
+const DEBUG_TO_FILE = false;
+const DEBUG_FILE_PATH = path.join(__dirname, 'debug.log');
```
StarDylan, line 22: These should probably be Environment variables, then we can enable/disable them without needing to change the code.

But instead we should probably also migrate to use a dedicated logging library (like [pino](https://github.com/pinojs/pino) since that seems to be popular). We could also potentially collect that timing data in production and display it in grafana!

But this is a good first step, and a good jumping off point to add additional debug tooling.

StarDylan (approved):

packages/server/src/services/gameWorld/serverWorld.ts
smbock42: I agree about the env variables and logging. Someone else already is apparently working on logging, but I'm not sure if it's going to have functionality similar to this. 

I figured since not many people are dealing with ticks it wasn't necessary to put in an env file, but lmk if you want changes

Issue #217, opened 2025-01-29
Optional Verbose or Muted Server Logs


ktaschek:
Currently, server logs are printed out to the console for various mob/world interactions. This makes the console cluttered, especially when trying to run cheat commands via the CLI (described in #187).

Add a debug mode that can be toggled to switch server logs on or off.

Deliverables:
- Debug mode that mutes/unmutes all logs in server console

rachelHartfelder:
Please update your issue to follow the guidelines found [here](https://github.com/sloalchemist/potions/wiki/Creating-Issues-in-the-Potions-Repository).

Issue #52, opened 2025-01-09
Chat Cut Off When on Edges of the Map


sreshtalluri:
When the user is on the edges of the map and chatting with other mobs, the messages should not be cutoff and should be entirely visible. Currently, the message is cutoff and the user is not able to read the full message (this happens when on the left and right most sides of the map).

Example:
<img width="399" alt="Screenshot 2025-01-08 at 1 38 58 PM" src="https://github.com/user-attachments/assets/2af45463-aa0b-41ec-82b9-402209959d05" />

Deliverables:

- Chat not cutoff and visible when on edges of the map

idavoong:
Please update issue to match guidelines listed [here](https://github.com/sloalchemist/potions/wiki/Creating-Issues-in-the-Potions-Repository)

PR #535, opened 2025-02-26
217 Optional debug logs for server & 52 Chat bubble cutoff bug fix

ethanhandelman:
Group: Andrew Estrada, Ethan Handelman, Isaac Lake

## Description
Created Server Logger, a singleton class responsible for handling sever logging to the console. It is designed to be used for all non-CLI logging. Allows for logging to be toggled on and off within the dev environment.
`packages/server/src/util/Logger.ts`

Moved speech bubbles to the right of the mob when they are on the far left side of the map, so player is always able to read the full message.
`packages/client/src/sprite/sprite_mob.ts`

All other changes involve replacing console.log calls with calls to the new Logger.

## Problem
Closes issue #217
Closes issue #52 

## Context

Alternatives for logger included third-party logging libraries or using a command on startup to turn logging on/off. This approach was chosen to maintain simplicity of the codebase and avoid any complications for other developers, as it is a simple change designed to make debugging and use of the CLI easier.

Alternatives for the bug fix could be to allow for users to view chat bubbles in the center of the screen by clicking on them, but this would be far more complex than necessary, and introduce problems with the UI.

This approach does not allow for redirecting of logs to logfiles.

There is no check for chat bubbles on the right side of the screen, but this only affects player chats as mob chats appear to the left by default. 

## Impact

 

The logging change does not directly break anything but will require an announcement in Slack to notify developers that all logging within Server should be redirected through Logger.

## Testing

Logger was tested both manually by toggling it on and off within the CLI while running the game and with additional unit tests added to `packages/server/test/cli/cheatHandler.test.ts`.

For the bug fix, testing was done by manually checking in game that chat bubbles appear to the right of mobs in the appropriate situations.

image
Example of chat bubble appearing on the right when it would have been cut off if on the left


StarDylan (changes requested):
A few notes. 

1. For next time this should've probably been 2 separate PRs since they are widely different issues (see [CONTRIBUTING.md](https://github.com/sloalchemist/potions/blob/main/CONTRIBUTING.md#atomic-pull-requests)).

2. Some things to consider for the logger, is that we probably want errors to always show up, even without debug logging enabled (this is where in other logging libraries we use log levels) and maybe warnings too. But it should be at least configurable to only see what you want (see https://stackoverflow.com/a/2031209). 

3. Also potentially in the future, in production, we might want to enable warnings and errors and save them elsewhere, which is really hard to do with the CLI interface, so we might want to use ENV vars as well to set the logging level.

Anyway, these are just some things to think about, feel free to msg me on slack to discuss it more. But in order to merge this PR I feel number 2 needs to be addressed.

StarDylan (approved):

PR #536, opened 2025-02-26
Prometheus 318

TylerNP:
Feat:
- Recongifured prom-client to push metrics to a pushgateway server
- Added regular polling 
- Added tick latency metric


StarDylan (approved):
Looks great

EthanV1920:
I want to add that they also added support for a heartbeat to an uptime dashboard that is now live at uptime.potions.gg

PR #543, opened 2025-02-26
Make Portals Work

logancostello:
## Description
We made the portals work. When the client clicks on the new world, an update is sent to supabase to set their world id to the new world. Then the page is reloaded, and the start page is skipped, automatically loading them into the world.

## Problem
This solves the issue of not being able to get between worlds

## Context
Another alternative was to somehow restart the game on the client side without reloading the page. We tried this extensively and could not make it work. In theory it'd be better because the client would not need to reconnect to the auth server, but we could not make it work. Maybe in the future.

## Impact
This change should not affect developers. There is some code in the repo (not from us) that does not allow for multiple servers to be run, but commenting it out seems to work. Also, I believe another team is fixing that bug.

## Testing
We used extensive manual testing for this. We did not test it with unit tests because the change heavily relies on communication between client, server, auth-server, and supabase.

https://github.com/user-attachments/assets/4390d93f-6372-4def-bb8b-a4350076e7ed

This video had much more lag than normal, but you can see how it works.


AndrewJMart (approved):
LGTM DYLAN U KNOW WHAT TO DO BRO REV THIS UP AND SHIP IT WHILE PRICES ARE LOW 
tlqq06gkyff21

jmdebarro (approved):
looks fabulous! 😍

packages/server/global.json
StarDylan, line 1: This seems to be an inadvertant build artifact, not sure why its not in the right location as defined in the `.gitingore`, but should be removed.

StarDylan (changes requested):
Small issue with left-in build data, but otherwise looks good.

StarDylan (approved):

PR #546, opened 2025-02-27
Affiliation check and stat boost

zarinayunis:
## Description
We added permanent stat boosts for when a player is affiliated with a species. We also ensured that a player's community was affiliated with the community of an item before allowing the item to be picked up. 

## Problem
There was no explicit problem with the way the game was designed before, however this feature was designed to make the game more creative. 

## Context
We contemplated creating a function for updating affiliation, however decided against it upon the realization that each players' stats changed successfully upon getting their affiliation. 

## Impact
Now, players are limited in what type of items they are able to pick up. The Affiliation Benefits section of the documentation should be updated. 

## Testing
We tested these changes manually in both the files playerAffiliationbenefits.test.ts and affiliatonBenefits.test.ts. In the first file, we ensured that 'pickup' was an available interaction if the community id of the player and the owner id of the item were the same. In addition, we checked to see if the item was a specialized attribute. In the second file, we ensured that each type of mob had their respective speed, attack, and maximum health as listed in the documentation.


packages/client/src/world/controller.ts
```diff
@@ -249,15 +249,24 @@ export function getPhysicalInteractions(
...
     interactions.push({
       action: 'pickup',
       item: item,
       label: `Pick up ${item.itemType.name}`
     });
   }
 
+  console.log(item.attributes['specialized_resource']);
+  console.log(ownerId);
+
```
StarDylan, line 269: We probably don't need these console logs

packages/server/src/mobs/mob.ts
```diff
@@ -909,6 +909,59 @@ export class Mob {
...
+                WHERE id = :id
+            `
+        ).run({ speed: newSpeed, id: this.id });
+        this.speed = newSpeed;
+        break;
+
+      case 'blobs':
+        // Blobs have higher attack (1.5x normal)
+        let newAttack = this.attack * 1.5;
+        if (newAttack <= 0) {
```
StarDylan, line 950: This should probably go in configuration, as the communities are also defined there, what if we want to add more effects?

StarDylan (changes requested):

Issue #258, opened 2025-01-31
Implement Individual-Based Ownership


chelseylearns:
### Summary:

We previously implemented community-based ownership, but there is now a need to support individual-based ownership as well. This will allow players to have personal ownership over items and structures, independent of community-based rules. 

### Goal:

Expand ownership system to include individual-based ownership while maintaining compatibility with existing community-based system.

### Deliverables:

- TBD

anvch:
Please follow the issue format as described here: https://github.com/sloalchemist/potions/wiki (i.e. format the description, add tags, assign yourself).

PR #549, opened 2025-02-27
feat: individual based ownership

nonetypes:
## Description
Implementation of individual based ownership to allow or disallow specified interactions based on who created the item.

Refactored item ownership logic to distinguish between community ownership and individual character ownership. Updated permissions handling logic in getPhysicalInteractions in client/source/world/controller.ts and validateOwnership in server/src/items/item.ts.

## Problem
Certain features, such as market stands, will require logic in place to determine ownership -- each individual player should only be able to perform certain interactions with items that they own or do not own. E.g., a player should not be able to collect gold from a market stand that they do not own.

Closes issue #258.

## Context
An alternative to the chosen solution which was considered was leaving in place strictly community based ownership, but making it hierarchical, i.e., to implement individual ownership, it would be necessary to make player characters a community of one. However, this may have been restrictive for future development, as it may have disallowed the ability for player characters to not join other communities.

## Impact
This change is not expected to break anything. Documentation may be necessary to make it more clear to other developers the method of specifying what individual interactions should be allowed or not allowed based on community ownership, character ownership, or non-ownership for specific items.

## Testing
Automated unit tests were written for both server side authentication of character ownership as well as client side UI interaction tests.


StarDylan (approved):
looks good!

PR #550, opened 2025-02-27
465 Message in a Bottle Read Interaction

### LINKED ISSUE #465: Add meaningful message and print message to in-game screen for Message in a Bottle Read Interaction.
Currently, when a message is read from a message in a bottle, a placeholder message is printed to the server terminal.

**Deliverables:** Change the placeholder message to something meaningful that builds upon our existing game vision. When a message is read from a bottle, the message should be displayed on the in-game screen.

annamakarew:
## Description
We changed the logic in handleSpeak(). Previously, it was coded to not show speech bubbles for a character's own speech. However, there is a precedent set that a character's own dialogue is shown in the game, so we removed that check from the function.

## Problem
Closes #465 

## Context
A pop-up that the user must close out of was considered, and printing the output to one of the tabs in the game was also considered. We chose using the speech bubble since it provides a temporary interface for the read message to be displayed. The tradeoff is that the message must be somewhat short since it must fit in the bubble.

## Impact
This change does not break anything, but could impact future developers who choose to use handleSpeak() to send a speak request from the server to client. 

## Testing
We tested the changes via manual testing. We tried a multitude of mixing chat options with reading the message to make sure that reading the message did not break any current game function.

image


packages/server/src/items/uses/read.ts
```diff
@@ -18,8 +18,8 @@ export class Read implements Use {
       return false;
     }
 
-    const message = 'some·message·here·(replace)';
-    mob.sendMessage(`You read the message: "${message}"`);
+    const message = `Greetings from the Deep… The sea whispers of forgotten treasures and ancient secrets hidden in the abyss. As you drift through the waters of Oozon, keep your eyes sharp. May the waves guide you.`;
+    mob.sendMessage(`${message}`);
```
StarDylan, line 22: Since there's only one bottle this is probably fine, but in the future we should abstract this a bit more.

StarDylan (approved):

PR #558, opened 2025-02-28
Improve Pathfinding When Target is Unwalkable

logancostello:
## Description
Updated the findNearestWalkableTile function to return targets closest to the player when there are multiple options for the closest tile to the original target.

## Problem
Before this changes, when you clicked on an unwalkable tile, you would pathfind to the closest tile to your target. However, when there were multiple closest tiles, you would always go to the rightmost one (due to how the breadth first search was implemented). Instead, when there are multiple options, the player should pathfind to the tile closest to them. 

## Context
Simply sorting the 8 neighbors by proximity to the player ensures that closest options are considered first. Sorting can be slow, but since the input is small, this change is not felt by the player. One option considered was looking at everything on the queue, then returning the smallest one, but that could be very slow if the queue gets large.

## Impact
No change for developers. They will enjoy better pathfinding when playing the game.

## Testing
Added an additional unit test. With the previous implementation, this test would fail because it would pathfind to the position to the right of the target, but now it finds the best one for the player

### DIFF
--- a/packages/common/src/pathFinder.ts
+++ b/packages/common/src/pathFinder.ts
@@ -383,11 +383,13 @@ export class PathFinder {
 
       // Sort by proximity to player, allowing for closest neighbors to be looked at first
       neighbors.sort((coord1, coord2) => {
-        const dist1 = Math.pow(start.x - coord1.x, 2) + Math.pow(start.y - coord1.y, 2);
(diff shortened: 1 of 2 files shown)


jmdebarro (approved):
lgtm

StarDylan (approved):
Looks good to me!

Issue #344, opened 2025-02-12
Make Fire World Interactable(Valoron)


jhuanny:
Create interactions between spirits and mobs.

Deliverables:

1. Items are interactable
- pickup obsidian --> then can build wall or partial wall
- pickup potion items

2. Connecting potion logic
- fire-flower = heart-beet-plant
- tar-pit = blueberry bush

PR #513, opened 2025-02-25
Refactor: remove asset clutter

clkodd:
See [510](https://github.com/sloalchemist/potions/issues/510)

- Duplicate `global.json` files deleted, centralized to world_assets directory
- `tsconfig.json` paths point to world_assets instead of client/static

PR #562, opened 2025-02-28
Valoron items spawn

emlai1263:
## Description
New items (Lightning bloom, tar, and gunpowder) now drop by their respective spawners (cloud, rock columns, and blaststone) in Valoron.

## Problem
This allows players who are associated with the fire world to craft potions. 
Furthers progress on issue #344 

## Context
Items are needed in each world to brew potions, and which items each world has was determined by the PMs

## Impact
Currently, the dev version's client can't access other packages, making testing virtually impossible without copying the respective atlas/global files into client. These shouldn't be pushed/merged, but were used to test our new implementations. Our changes should be a rough outline on what needs to be changed once the json issue is resolved.
Other devs who are working with global.json, global-atlas.josn should see small addition/changes, but it should not break other worlds.

## Testing
We tested this by manually entering the world and verifying that the correct spawners spawned as expected.

image
image


Twicee:
Added more assets and spawn logic for firewold. #513 broke prod because it now causes us to reference assets outside of the server package in setup.ts. However once this bug is fixed our code should work fine as we have tested locally and ensured its functionality

StarDylan (approved):
Looks good!

PR #563, opened 2025-02-28
fix: Made portal button not appear for worlds down or for current world

### LINKED ISSUE #554: Only Show Available Worlds to Travel To
Currently when you look at a list of worlds to travel to in the portal, the current world shows up, which is not needed. 

Deliverables: 
Only show the worlds you can actually travel to.

jmdebarro:
## Description
When clicking "Enter Portal" no longer shows worlds that are down, checks health of server, and does not show current world.

## Problem
1. Doesn't show current world which you wouldn't need to tp to as you are on
2. If user were to tp to world that is down, there would be no way to change their world id in the production database back to a world that could host them. Essentially, that character is lost until the world comes back up.
3. Closes #554 

## Context
Not checking world health status. 
I have hardcoded the world-names to their IDs on the uptime service. Do no believe this is a problem as no new worlds will be added and our worlds on that service will not be edited or deleted.

## Impact
No breaking

If developers are trying to change world locally, they would be able to if they ran the servers themselves, but now, if a user is running a server locally, and attempts to run another server and click the portal, despite the server being up locally, it would determine whether or not to show the portal option based on production server.

## Testing
Manually tested changes, read the network traffic and used logs to see if requests are working and delivering the expected results

Previous, show every option
old

Current, show available options (servers are up)
notcurrent

Current when no servers are up. Production is down currently
EverytingDown

### DIFF
--- a/packages/client/src/scenes/portalMenuScene.ts
+++ b/packages/client/src/scenes/portalMenuScene.ts
@@ -7,14 +7,14 @@ import { stat } from 'fs';
 export class PortalMenuScene extends Phaser.Scene {
   constructor() {
     super({ key: 'PortalMenuScene' });
-    }
-    
-    world_map = new Map<string, number>([
-        ['fire-world', 3],
-        ['water-world', 4],
-        ['test-world', 5]
-      ]);
-    
+  }
+
+  world_map = new Map<string, number>([
+    ['fire-world', 3],
+    ['water-world', 4],
+    ['test-world', 5]
+  ]);
+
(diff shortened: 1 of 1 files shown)


logancostello (approved):
Man this is inspirational. Truly a legend. Might have just fixed the biggest bug in the game with such a simple solution. Makes a man want to cry

StarDylan (approved):
I don't love having the local dev experience dependent on the uptime website. I'd much rather it reach out to the world directly to see if its up. But that might be a later issue

PR #564, opened 2025-02-28
497 market stand sell any one item

AkashPraveen:
## Description
Fixed the Market Stand Code to reflect the single-item market stand set-up that we agreed on with PM, allowing players to sell one type of item at a time at a market stand.

## Problem
The existing database architecture did not support selling multiple types of items from a single stand. The database SQL logic was structured around storing strings/numbers, not full objects, making it difficult to introduce multi-item inventories without significant structural changes to the database schema and game logic. Professor Pierce recommended we abandon that approach and renegotiate the functionality

## Context
After discussing with the PM, we explored alternatives such as allowing the market stand to have multiple item slots (e.g., selling three different items at once). This would work with the current database logic, but be messy and not an ideal solution. Instead, we opted for a simpler solution where the stand can hold only one item type at a time, and once empty, a new type can be added. This balances flexibility and minimal engineering overhead while maintaining usability.

## Impact
- Does not break existing functionality or require changes to other systems.
- No direct impact on other development teams.
- Preserves compatibility with current game mechanics while providing a scalable approach for future improvements.

## Testing
- Existing test cases for basic market stand interactions confirm this implementation works as expected.
- Since the logic is similar to the potion stand, the current test structure remains valid.
- Additional tests will be added as the market stand functionality expands.


StarDylan (approved):
Looks good! Thanks for filling me on the details of your solution, it was very helpful

Issue #559, opened 2025-02-28
setup.ts references files from outside of server package


clkodd:
A change in [513](https://github.com/sloalchemist/potions/pull/513) has `setup.ts` referencing files directly from the world_assets directly, outside of the server package. 

Deliverables: `setup.ts` will reference assets using [potions.gg](potions.gg) links.

PR #567, opened 2025-03-01
559 setup.ts references files from outside of server package

clkodd:
## Description

Changes paths in `setup.ts` and `buildWorld.ts` to potions.gg URL instead of world_assets.

## Problem

Closes [559](https://github.com/sloalchemist/potions/issues/559)

## Context

Client already imports via URL, so this changes server files to import via URL as well.

## Impact

 

Fixes an existing issue. Should not affect others.

## Testing

Used Render preview, tested manually on dev and prod.


StarDylan (approved):
Looks good!

PR #569, opened 2025-03-01
feat: Added Community Favorabilites to Player Info UI

### LINKED ISSUE #568: Seeing Favorability UI
Currently, users cannot specifically see their players' favorability numbers with the other communities. However, the player should be able to keep track of their changing favorability with each community whenever they fight and give items. 

Deliverables:
-   Allow the player to see the favorability they have with each community.

Andrew76xx:
## Description

Added a new feature where the user (alchemist) can see what favorability they have with each community (blobs, silverclaw, fighters, and future communities)

## Problem

Closes #568 

## Context

We implemented this where the player can press the **_F key_** and the numbers will appear. Originally, we believed that it should be in the Player tab, but we recognized that it will make more cluttered than it should be. We have chosen this approach as we believe that it should be easily found, in case the player wants to keep track of favorability when doing some type of action. Like what is done with the potions handbook, this reduces the amount of tabs in the UI. 

A new attribute, favorabilities, was added into the `mob.ts` in the `common` folder for this implementation.

## Impact

 

This change added a new parameter for mobs so it affected some previous test cases which has been resolved by us.

## Testing

Because this is mainly a change in the UI, no tests were made.

Result:
image

### DIFF
--- a/packages/client/src/scenes/uxScene.ts
+++ b/packages/client/src/scenes/uxScene.ts
@@ -65,6 +65,7 @@ export class UxScene extends Phaser.Scene {
   defenseText: Phaser.GameObjects.Text | null = null;
   speedText: Phaser.GameObjects.Text | null = null;
   affiliationText: Phaser.GameObjects.Text | null = null;
+  favorabilitiesText: Phaser.GameObjects.Text | null = null;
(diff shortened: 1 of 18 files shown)


StarDylan (approved):
Looks good!

Issue #570, opened 2025-03-02
List of Keybinds Features


Andrew76xx:
Though there are keybinds in the game, the user has no access to what those keybinds are and what they do. The user should be given a guide on what keybinds the game has.

Deliverables:

Allow the player to see the available keybinds in the game.

PR #572, opened 2025-03-02
570 keybind manual

Andrew76xx:
## Description

Added a feature to allow the player to see existing keybinds in the game.

## Problem

The issue is that though the keybinds are implemented, the players has no idea what keybinds are working and what they do. We believed that it will be helpful if there is a guide provided

Closes: #570 

## Context

We implemented this where the player can press the **_K key_** to see all existing keybinds. We have chosen this approach so that there will not be any extra tabs. This is done similarly to the handbook and favorability implementation.

A new container was added into `uxScene` file. Added text into the Player Tab to let players know a manual exists

## Impact

 

Should not affect other developers since only the `uxScene` file is modified.

## Testing

Because this is mainly a change in the UI, no tests were made.

Screenshot1

Screenshot1


StarDylan (approved):
Looks good! Good thinking about adding the help text to the player tab

PR #591, opened 2025-03-03
529 update potion effect tests

### LINKED ISSUE #529: Update Potion Effect Tests
Current potion effect unit tests are lacking and a bit confusing. We want to add more tests to cover a wider range of edge cases and update the current tests.

Deliverables:
* New unit tests for each potion effect
* Updated versions of the current tests

charm-13:
## Description
Updated potion effect tests.

## Problem
Closes #529. Previous potion effect unit tests were lacking and a bit confusing. We added more tests to cover a wider range of edge cases, updated the previously implemented unit tests, and reorganized the potion effect tests to have each type in a different file all under the `potionEffects` directory.


StarDylan (approved):
Looks good

PR #595, opened 2025-03-04
publish chat messages to clients

jqueja:
## Description
Created another ably channel for chat messages to display chat bubble when client receives messages.

## Problem
Players will now be able to communicate via text chat.

## Context
Before this change, the message was not being sent anywhere to publish to other clients. It was just showing up on a singular client's side.

## Testing
Testing not required.
To have confidence that it worked with other clients, we opened up an incognito tab to simulate multiple clients subscribing to the ably channel.

image


StarDylan (approved):
Looks good

PR #597, opened 2025-03-04
fix: log error rather than crash when mob not found on persistence

jackalnom:
## Description
World persistence blows up right now if the player dies at an inopportune time.

## Problem
Patches the issues above to not crash. I'm logging an error instead and I abort from the persistence method.

## Context
This persistence & presence code is not great. I'm sure there is a better solution but trying to get the server to stay up.

## Impact
Server stops crashing.

## Testing
I retriggered the error and it no longer crashes. Seems to work much better now.


logancostello:
I'll be tackling this problem soon.

StarDylan (approved):

PR #598, opened 2025-03-04
Made shipwreck transparent when character walks behind

0felicia0:
## Description
When the character walks behind the shipwreck, they will be visible as the ship becomes translucent. A coral was also removed that was at the same spot as the ship.

## Problem
Visibility of the character

## Context
Visibility of the character

## Impact
No impact should occur as this is a world asset specific change

## Testing
Dev environment

image


halliec6 (approved):

Parshana007 (approved):

packages/client/src/scenes/worldScene.ts
```diff
@@ -535,6 +536,28 @@ export class WorldScene extends Phaser.Scene {
...
       });
+
+      Object.values(world.items).forEach((shipwreck) => {
+        const ship = shipwreck as SpriteItem;
+        const heroX = Math.floor(x);
+        const heroY = Math.floor(y);
+
+        // Calculate the width and height in tiles
+        const shipWidthTiles = 110 / 32;
+        const shipHeightTiles = 115 / 32;
```
Parshana007, line 547: Consider changing 32 to use TILE_SIZE

PR #608, opened 2025-03-04
583 dynamic global data loading

athapar9:
## Description

This PR refactors the server to fetch the global.json configuration dynamically from a from https://potions.gg/world_assets/global/server/global.json rather than a static import in setup.ts and buildWorld.ts.

## Problem

Previously, we used a statically imported global.json file. However, this requires manual updates and isn't scalable so we want to instead dynamically load the necessary data, allowing the system to automatically stay up-to-date with the latest version from the server.

## Context

An alternative approach would have been to continue using the static import and manually update it with every configuration change. However, dynamically fetching the configuration at runtime ensures that the server always uses the latest data, avoiding the need for manual updates. The trade-off of this approach is introducing asynchronous loading.

## Impact

 

This change does not break any existing functionality. 

## Testing

Manually tested.


StarDylan (approved):
Looks good

PR #612, opened 2025-03-05
Add Zod Schema Validation to global.json Configuration File

### LINKED ISSUE #602: Add ZOD schema validation to global.json config file at build time
# Problem

The client and server both trust that their generated `global.json` files aren't malformed so its important that its impossible for us to write broken `global.json` files to them. Currently we are blindly copying `packages/json-generator/global.json` to the client and server with no guarantees that the contents will be in any specific shape.

Note that `packages/json-generator/global.json` is written to both `world_assets/global/client/global.json` and `/world_assets/global/server/global.json` at build time for the client and server respectively. 

Here you can see how we're currently unsafely parsing `global.json` in `packages/json-generator/src/generator.ts :

```typescript
if (val == 'client') {
    const rawJson = readFileSync('./global.json', 'utf8');
    const json_client = JSON.parse(rawJson);
    client_breakup(json_client);
    writeFileSync(
      '../../world_assets/global/client/global.json',
      JSON.stringify(json_client, null, 4)
    );
    ran = true;
  } else if (val === 'server') {
    const rawJson = readFileSync('./global.json', 'utf8');
    const json_server = JSON.parse(rawJson);
    server_breakup(json_server);
    writeFileSync(
      '../../world_assets/global/server/global.json',
      JSON.stringify(json_server, null, 4)
    );
    ran = true;
  }
```
# Solution

To resolve this, we need to add schema validation to the `packages/json-generator/global.json` file at the time of distribution inside of `packages/json-generator/src/generator.ts`. This way, we will fail early on schema validation errors, which will be caught by our CI pipeline instead of realizing salient errors at runtime.

AdamMikacich:
## Description

> This is a devops task with the goal of making configuration files less error prone

The client and server both trust that their generated global.json files aren't malformed, so its important that its impossible for us to write broken global.json files to them. Before this, we were blindly copying `packages/json-generator/global.json` to the client and server with no guarantees that the contents would be in any specific shape. This PR adds schema validation with `zod`, ensuring that the generated files come from a `global.json` file that conforms to  the expected structure before distribution.

## Problem

Closes #602

## Context
- **Generated `global.json` in `.gitignore`:** Since the client and server `global.json` files are generated at build time, they should not be version controlled.
- **Zod schema for validation:** Ensures `packages/json-generator/global.json` always matches the expected data shape. If the shape is incorrect, validation fails early.
- **Snapshot tests to validate generation:** Tests ensure generated client and server files match the old (pre-schema) output. If you modify the source `global.json` shape, you must update `expectedClient.json` and `expectedServer.json` accordingly for the tests to pass.

## Impact
- **Early failure on invalid JSON:** CI will catch malformed files before they impact the client or server at runtime.
- **Developer workflow:** If `global.json` changes, developers also update the corresponding `expectedClient.json` and `expectedServer.json` tests. This avoids surprises in production and keeps everyone aware of schema changes.

## Testing
- **Automated Tests:** Added Jest tests in `packages/json-generator` to ensure that:
  1. The files generated by `generator.ts` match expected outputs (`expectedClient.json` and `expectedServer.json`).
  2. The schema validation with `zod` passes with valid data and fails as expected on malformed JSON.
- **Manual Verification:** Confirmed locally that modifying `global.json` leads to test failures if the shape doesn’t match the schema.


StarDylan (approved):
Looks great! Hopefully will cut down on downtime

Issue #578, opened 2025-03-03
Keybinds for common interactions


joshualeeee:
Deliverables: keybinds for pickup, stash, drop

PR #619, opened 2025-03-05
Keybinds for common interactions. Issue #578

apervmm:
## Description
Adding a short cut keyboard interactions on top of the WASD including and not limited to:
1. Picking up item with keyboard on <e/E> 
2. Dropping item with keyboard on <e/E> (if item is on hands)

## Problem
In the current state of the game, it becomes difficult to toggle actions using on game buttons, so adding a feature to play the game using the keyboard adds smooth and intuitive user interaction. 

## Context
Initial thought was to have a separate **Keyboard** buttons for each interaction, but from UX perspective it would be harder for the user to remember all possible hotkeys, so we used **Apple's Home Button** idea to toggle the same button for different tasks depending on the state of the process.

## Impact
1. This might require a HotKey Documentation update
2. During the client side testing, it will be easier and faster for developers and clients to interact with objects.
3. Not necessarily to notify, but great if they all know of this feature.

## Testing

1. As of right now, there were only manual testings of interacting with objects through HotKeys under different conditions.

### DIFF
--- a/packages/client/src/scenes/worldScene.ts
+++ b/packages/client/src/scenes/worldScene.ts
@@ -53,7 +53,13 @@ export class WorldScene extends Phaser.Scene {
   terrainWidth: number = 0;
   terrainHeight: number = 0;
   nightOpacity: number = 0;
-  keys: { [key: string]: boolean } = { w: false, a: false, s: false, d: false, e: false };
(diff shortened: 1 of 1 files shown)


StarDylan (approved):
Looks great

PR #620, opened 2025-03-05
560 implement logging library for server

### LINKED ISSUE #560: Implement Logging Library for Server
Deliverables:

- Addition of logging library to server package
Should be able to route logging to logfiles and change log level according to .env vars 

- Reroute all current logging within server to new logging library

ARES-79:
## Description

We added the logging library log4js-node for the server.
As part of this process we:
- updated the logger.ts file to have a log4js logger for appending to a file and to the console,
- updated the CLI to make use of logging levels, 
- updated logging instances that were invalid for the logger (didn't have a sting type first argument)

## Problem

This solves the issue of a crowded server console when trying to use the CLI by allowing to change the log level of the console while still having access to all the logs within a log file. 
Closes #560 

## Context

The logging libraries of pino and winston were both considered as alternatives to log4js but log4js was chosen for its ability to adjust log level at runtime which we wanted to make use of for the CLI. 

The log files are saved to 'packages/server/logs' for easy access that is isolated from code.
The logger is configured to save a maximum of 3 log files that can have a max size of 10MB. This is so there is a decent amount of logging history across multiple sessions without storing an excessive amount in each developer's files. 

## Impact

 

This is not a breaking change since developers can still use console.log() but they should be notified that they should be using the logger now instead. It also impacts developers because they must now provide a first argument that is a string (meaning you can no longer log just an error object or just a blob object). While this may be a small burden for developers, we do think there is significant value in ensuring there is a message with these objects for more usable log messages. 

## Testing

Two unit tests replaced the old logging tests in the `packages/server/test/cli/cheatHandler.test.ts` to ensure that the logic for handling CLI commands for adjusting the log level was correct.
No further unit tests were added since they would be testing the functionality of an external library.
We performed manual sanity testing to ensure that the logging library behaved as expected by checking the console output and the log file output.


StarDylan (approved):
Looks great! Hopefully this helps the dev experience a bit

Issue #622, opened 2025-03-05
Reduce NPC caps and capture/visualize tick metrics


smbock42:
Reduce NPC caps to reduce clutter and mob slowdown of server. 

Graph and capture tick time and mob complexity per tick

PR #628, opened 2025-03-05
Reduce Mob Caps and Create Metrics collection for mob ticks and tick time. 

smbock42:
### Description

Implemented a Game World Metrics Dashboard with charts for tick time and mob metrics. Added metrics logging for mob and tick processing. Adjusted mob caps and spawn rates for more balanced gameplay.

### Problem
Mob tick time is taking too long and gets longer with more mobs joining. We are diagnosing where this is going wrong. 

### Context
- Using data to see how tick time scales with amount of mobs. 

### Impact
- The changes do not break existing functionality but introduce new files and logging mechanisms.

### Testing
No tests needed - just added functionality. 

### Screenshots (if appropriate)
Before mob cap reduce
image (1)
after mob cap reduce
image (2)


smbock42:
Linking #622

packages/server/src/services/gameWorld/serverWorld.ts
```diff
@@ -19,16 +19,102 @@ import path from 'path';
...
+};
+
+function debugLog(message: string, data?: Record<string, unknown>) {
+  const logEntry: LogEntry = {
+    timestamp: Date.now(),
+    type: 'debug',
+    message,
+    data
+  };
+
```
StarDylan, line 105: I'd change this out with the standard logger defined in #560  / rename it so it's clear what each logger does.

StarDylan (changes requested):

packages/server/src/services/gameWorld/serverWorld.ts
StarDylan: This comment still applies, can you please address it?

StarDylan (changes requested):
Please see my comment

packages/server/src/services/gameWorld/serverWorld.ts
smbock42: @StarDylan resolved

StarDylan (approved):

PR #637, opened 2025-03-05
Bug: Added checks so we never check attributes for dead mobs

### LINKED ISSUE #586: Fix: characters being searched for that do not exist in the database
In mob.ts, get poisoned() sometimes returns an undefined value from the database, meaning that we are searching for a mob that does not exist in the database. 

Deliverables: All returns from the database in get poisoned() should return a valid mob, and the error message should never be reached.

sbryukho:
## Description
Added checks to see if dead mob is no longer in the database within files potionEffects, mob, flee, heal, and hunt to avoid referencing undefined attributes.

## Problem
closes #586 

## Context
Had to be patched - reference #585 fix.

## Impact
No documentation required, will not affect developers.  

## Testing
No formal testing necessary, just ran the game and checked if the checks appeared in console log.

image
image

### DIFF
--- a/packages/server/src/mobs/mob.ts
+++ b/packages/server/src/mobs/mob.ts
@@ -184,7 +184,10 @@ export class Mob {
       SELECT health FROM mobView WHERE id = :id
       `
     ).get({ id: this.id }) as { health: number };
-
+    if (!mob) {
+      console.error(`Get Health: Mob with id ${this.id} and name ${this.name} not found`);
+      return 0;
+    }
     return mob.health;
   }
 
@@ -196,8 +199,8 @@ export class Mob {
     ).get({ id: this.id }) as { poisoned: number };
 
     if (!mob) {
-      console.error(`Mob with id ${this.id} not found`);
-      return 0; // Return a default value or handle it differently
+      console.error(`Mob with id ${this.id} and name ${this.name} not found`);
+      return 0; 
     }
     return mob.poisoned;
   }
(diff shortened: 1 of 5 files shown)


packages/server/src/mobs/mob.ts
```diff
@@ -435,6 +440,10 @@ export class Mob {
   }
 
   changeHealth(amount: number) {
+    if (!this || !Mob.getMob(this.id)) {
+      //console.error(`${this.name} is no longer valid or does not exist in the database.`);
```
StarDylan, line 444: Maybe convert this to logger defined in #560

packages/server/src/mobs/mob.ts
```diff
@@ -655,7 +665,7 @@ export class Mob {
 
     if (this.poisoned == 1) {
       const deltaDamage = Math.floor(Math.random() * -10);
-
+      //console.log(`changing ${this.name} health`)
```
StarDylan, line 668: Same here

packages/server/src/mobs/mob.ts
```diff
@@ -1024,12 +1034,22 @@ export class Mob {
     this.updatePosition(deltaTime);
 
     if (this.type !== 'player') {
+      // Check if mob exists and is valid
+      if (!this || !Mob.getMob(this.id)) {
+        //console.error(`${this.name} is no longer valid or does not exist in the database.`);
```
StarDylan, line 1039: And here

StarDylan (changes requested):
I would go through and convert all the console.logs to the logger defined in #560 as errors or warnings (depending on severity). Also uncomment all the logging statements

StarDylan (commented):
Sorry, I meant all the console.log / console.errors in your code should be converted to logger statements, not just the ones that I pointed out. Also there are a some lines where the console.log is still commented out, it should be removed or converted to the logger

StarDylan (approved):
Looks good

PR #643, opened 2025-03-05
feat: Unstash swaps item in inventory with item in hand

### LINKED ISSUE #633: Unstashing item from Inventory swaps with item in hand
### Problem
Currently, when clicking on an item in the Inventory tab, this item is "unstashed," which effectively drops it.

### Fix
Instead, our feature swaps this item in our inventory with the item in our hand. To drop an item in the Inventory, simply place it in your hand, then drop it using the "drop" action in the "Actions" tab.

xgarcia17:
## Description
Unstashing an item in your inventory (by clicking the item in the "Inventory" tab) now swaps that item with the item you are carrying in your hand. Before, unstashing an item would drop it onto the ground at your feet.

## Problem
This version of "unstash" aligns with what is the standard for many other games when clicking on an inventory item. It also makes life easier for the user to be able to equip the item in their inventory when they want instead of dropping the item.

Closes #633 

## Context
The old design dropped items in your inventory, but now we swap it with what is in your hand. These changes were made as step 1 to our next issue, which is allowing items in your inventory to be directly added to a cauldron when trying to "add ingredient" to the cauldron.

## Impact
This change shouldn't negatively impact developers. It should be noticeable to developers, so we don't think developers will need to be alerted.

## Testing
We tested by playing the game and playing around with our inventory.

**Note:** These screenshots do demonstrate the functionality of our changes, however, because the client would lose focus when we screenshot our browser, each screenshot is from a different instance of the game.

**Inventory:** Sunflower, **In hand:** Log
image

_**Click 'Sunflower' in inventory tab**_

**Inventory:** Log, **In hand:** Sunflower
image


StarDylan (approved):
I think that makes a lot of sense, thanks!

PR #645, opened 2025-03-05
[Bug Fix] Update Wall Building to only allow Closer Walls

MorrowCameron:
## Description
Lower the maximum range of finding valid locations for item placement (drops, wall builds, etc.)

## Problem
Currently walls are being placed far away from the playable character causing confusion if nearby areas already have items.

## Context
Lowering the maximum range was chosen due to the easily reversible nature of a parameter change with added functionality to destroy dropped items if they could not be placed. This allows the change to scale with just a parameter change in the future and not abandon items.

## Impact
No external dev impact, the range can be increased as a parameter if needed by other developers.

## Testing
- Manual Testing
- Unit Testing


NickPerlich (approved):
Walls no longer spawn in random locations when no valid placement location is found.

StarDylan (approved):

PR #648, opened 2025-03-05
Fix Player Not Found Bug When Updating Player State 

logancostello:
## Description
When a player leaves the game (or loses focus), the server updates the player data in the DB and then removes the player from the world. The server awaits the update in order to ensure the order is correct.

## Problem
Previously these two actions had no guaranteed order, resulting in a race case where the player upload would crash the server since it was trying to upload information from a player that was not in the world. This is a more permanent solution to #597

## Context
This approach was chosen give the simplicity of it. This approach seemed obvious, so no others were considered.

## Impact
No impact on developers. Servers should stop crashing when the game loses focus.

## Testing
I manually lost focus 25 times, and the race case error did not hit. This, plus the logic of it, makes me think it has been solved for.

### DIFF
--- a/packages/server/src/services/clientCommunication/ablyService.ts
+++ b/packages/server/src/services/clientCommunication/ablyService.ts
@@ -88,7 +88,7 @@ export class AblyService implements PubSub {
       this.broadcastReloadPageTrigger();
     });
 
-    this.broadcastChannel.presence.subscribe('leave', (presenceMsg) => {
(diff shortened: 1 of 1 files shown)


jmdebarro (approved):
makes sense lgtm

AndrewJMart (approved):
LGTM

StarDylan (approved):

PR #654, opened 2025-03-06
653 certain terraform errors arent being caught by custom error handling

### LINKED ISSUE #653: Certain terraform errors aren't being caught by custom error handling
There are issues that can occur with terraform that don't have custom error messages.

NickPerlich:
## Description
Added new custom error messages for various errors related to tf.ps1 and tf.sh scripts 

## Problem
This is important so that developers unfamiliar with the terraform workflow can get back to work faster instead of having to worry about debugging. 
"Closes #653 " 

## Context
No alternatives were considered as this method has worked for previously handled errors. 
Extra messages were added so there was no change to the design or tradeoffs of the current method of error handling. 

## Impact
This change does not break anything and does not require documentation because it essentially is documentation.
This change should make it easier for developers to get to the point of being able to work on the project faster.
Nobody needs to be notified.

## Testing
Changes were not tested beyond checking to see that the custom error message pops up when the error is intentionally created
This change does not need to be tested as nothing new was added and the system already worked. It was simply made to catch more errors.

### DIFF
--- a/tf.ps1
+++ b/tf.ps1
@@ -64,6 +64,9 @@ if ($exitCode -ne 0) {
     if ($output -match "The following organization members have reached their maximum limits") {
          Write-Host " You have more than 1 active Supabase project in your organization." -ForegroundColor Red
     }
+    elseif($output -match "Ably API token cannot be an empty string.") {
+         Write-Host "Make sure you have followed the README instructions to create your tf.tfvars file."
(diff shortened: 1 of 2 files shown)


MorrowCameron (approved):

StarDylan (approved):
Looks great!

Issue #649, opened 2025-03-06
Create World Script Only Succeeds When AuthServer Is Running


logancostello:
When you run 'pnpm run create world-name' it only works when the auth server is running. This is because the startup of a world calls getWorlds(), which is an async request to the auth server. This should not be included on startup.

Deliverables:
Make startup work when auth server is not running.

PR #656, opened 2025-03-06
fix: lazily load worlds from auth server at run time to prevent error in "pnpm run create <world>"

jmdebarro:
## Description
Instead of populating worlds in the constructor of enter portal class, I populate them when interact is called

## Problem
#649 
All interaction items are called and their constructor is called during "pnpm run start <world>" causing an error to appear as getWorlds() is called which requires auth-server connection. This error is confusing and can be avoided.

## Context
I was going to make the interact ufnction async and await the populateWorlds() function but that causes issues down the line as you must return a boolean, not a Promise.

## Impact
doesn't break anything, just removes error during create script

## Testing
Manually tested to see if portal would still show worlds
Removed tests that dealth with old constructor. Tests still with interactions as interactions are not changed.

previous, error on build
bad

current
good

### DIFF
--- a/packages/server/src/items/uses/enterPortal.ts
+++ b/packages/server/src/items/uses/enterPortal.ts
@@ -13,7 +13,6 @@ export class EnterPortal implements Use {
   constructor() {
     this.key = 'enter';
     this.worlds = [];
-    this.populateWorlds();
   }
 
   private async populateWorlds() {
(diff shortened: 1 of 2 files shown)


AndrewJMart (approved):
LGTM

logancostello (approved):
lgtm

StarDylan (approved):

PR #664, opened 2025-03-06
Fix inventory dropping

gabrielriedel:
## Description
I fixed the logic in the 'unstash' function in both the server and client so that when you unstash an item from your inventory, it does not drop it on the floor but instead moves that item to your hand (carrying). If you are currently carrying an item, and you want to unstash another item from your inventory, the function first stashes the item you are carrying then moves the other item from your inventory to your hand.

## Problem
According to the PM's, the previous logic of unstash dropping items on the floor was incorrect and needed to change.

## Context
I did not really consider many other alternatives; this problem seemed cut and dry. Instead of dropping the item, it needed to be added to the mob's hand, so that needed to be reflected in the image on the client side and the database on the server side.

## Impact
This change does not break anything else and is highly decoupled from any other code logic other than the broadcasting of the unstash message, but nothing in that message sending structure has changed.  

## Testing
I tested both manually and automatically. In both cases I went through workflows of picking up an item, stashing it, unstashing it, and seeing the user carry it once again.

### DIFF
--- a/packages/client/src/world/item.ts
+++ b/packages/client/src/world/item.ts
@@ -121,10 +121,12 @@ export class Item extends Physical {
     if (!mob.position) {
       throw new Error('Mob has no position');
     }
-    this.carried_by = undefined;
-    this.position = position;
-    world.addItemToGrid(this);
     world.removeStoredItem(this); // Remove from stored items
+    if (mob.carrying) {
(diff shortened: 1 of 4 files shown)


StarDylan (commented):
I think this was already done by #643, correct me if I'm wrong

Issue #587, opened 2025-03-03
Market Stands Do Not Accept All Items to be Sold


AkashPraveen:
Market stands only allow certain items to be sold for whatever reason. Some items cannot be put into the market stand. 

Deliverables: 
- Make it so any item can be sold at the market stand

Goal: Make sure all items can be sold at the market stand.

PR #666, opened 2025-03-06
Market Stand Interactions Fixed

AkashPraveen:
## Description
There were existing mistakes with the interactions for the market stand that were built upon when we switched to the single item at a time functionality. I went through and renamed and redid a few of those interactions.

## Problem

Closes: https://github.com/sloalchemist/potions/issues/587

## Context
There was some old interaction functionality that we had from previous iterations that we could have reverted to to try to solve this bug, but we decided that code may cause conflict errors with the new system and it would be easier to rewrite it.

## Impact
Should not break anything else and does not interact with any other team's work.

## Testing
Existing Potion Stand Tests and Market Stand Testing Suite cover this.


StarDylan (approved):
Looks good

PR #669, opened 2025-03-07
Fixed weakened green potion effect

### LINKED ISSUE #634: Fix Weakened Green Potion Effect
The weakened effect for the green potion is supposed to cause less damage to poisoned mobs over a shorter duration. Currently, mobs are not poisoned at all, but the "effect" from the potion lasts for the shorter period of time. 
When a player drinks a weaker green potion, we want the mobs that attack them to be poisoned to a lesser effect for a shorter period of time.

sgill33:
## Description

The weakened green potion (when mixing a potion similar to but not the same as a green potion) did not have any effect. A player's damageOverTime value was reduced, but the reduced value was not used anywhere. We slightly refactored the green potion logic so that a weakened green effect is possible where is doesn't do as much damage and it runs out quicker.

## Problem

Closes #634 

## Context

We needed to turn the damageOverTime and poisoned so the weakened green effect was possible. `potionEffects.ts` was setup to have a weakened potion effect, but other files weren't so it was doing nothing.

## Impact

 

- The values of damageOverTime and poisoned were changed to reals instead of ints to allow them to act as multipliers
- If you make a potion similar to a green one, it will poison and enemy, but not as much and for not as long

## Testing

added a weakened green potion effect test to `greenPotion.test.ts`


StarDylan (approved):
Looks good!

PR #678, opened 2025-03-07
Fix: Cauldron shouldn't blow up with bomb potion

### LINKED ISSUE #673: Cauldron Shouldn't Be Able to Be Blown Up With Bomb
We destroy everything in the bomb radius, including cauldron which is bad because we can't spawn it back in.

Deliverables
- Make cauldron immune from Bomb

jacksonisidor:
## Description
Previously, the bomb potion blew up *everything* in a 3 tile radius besides the mob that drank it. Now, it spares the cauldron.

## Problem
The cauldron will be gone forever if someone blows it up on prod, which kinda defeats the purpose of the game when you can't make a potion anymore. Closes #673 

## Context
The cauldron is a necessary staple of the game, so a player should not be able to remove it from the world in any way. We decided to make it indestructible by having the bomb potion logic just skipping cauldrons as it goes through the effects. Another option considered would be to add something to the cauldron class itself making it indestructible to *anything*, even non-bomb potion threats. However, this is the only feature of the game that could currently affect a cauldron, so that approach wasn't quite necessary yet. 

## Impact
This change shouldn't break anything or affect any developers. 

## Testing
We already thoroughly tested the bomb potion in last weeks PR, so I just adapted it to make sure that cauldrons are still in the world after being in close proximity to the bomb.

### DIFF
--- a/.github/workflows/client-deploy.yml
+++ b/.github/workflows/client-deploy.yml
@@ -51,6 +51,7 @@ jobs:
       - name: Copy Static Files
         run: |
           cp -r ./packages/client/static ./packages/client/dist/static
+          cp -r ./packages/client/world_assets ./packages/client/dist/world_assets
           cp -r ./packages/client/githubPages/index.html ./packages/client/dist
 
       # Step 8: Deploy to GitHub Pages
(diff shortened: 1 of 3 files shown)


StarDylan (approved):
Looks great! Yeah destroying the cauldron would kinda defeat the purpose of the game.

Issue #646, opened 2025-03-05
Fix Leaderboard UI


DarrenH6293:
Having a large amount of gold would result in the gold text overlapping with the username in the Leaderboard UI.

Deliverables:
- Make sure the gold text does not overlap with the names in the Leaderboard UI.

PR #681, opened 2025-03-08
Fix Leaderboard UI

DarrenH6293:
## Description

Added a function in `leaderboardScene.ts` to format the gold text to be shorter when exceeding a certain amount. 1,000 to 1K and 1,000,000 to 1M. 

## Problem

#646 
If the gold amount is large enough, the text would overlap with the username in the leaderboard UI causing readability issues.

## Context

Alternatives were making the leaderboard UI bigger, but we don't want that since we already introduced a change to make it smaller before. Therefore we chose to format the gold text amount to make it concise.

## Impact

 

This does not break anything or affect other developers.

## Testing

No automated tests, could be manually tested by giving yourself gold and seeing the Leaderboard UI.

**Before:**
image

**After:**
image
image


StarDylan (approved):
Looks great!

PR #682, opened 2025-03-08
Fix mob aggro logic

DarrenH6293:
## Description

Modified the threshold for when hungry mobs attack in `hunt.ts` to match with the game wiki.

## Problem

#647 
 The hunger threshold for attacking was different than what was stated in the wiki.

## Context

The mob aggro behavior should match with the game wiki to prevent confusion for players.

## Impact

 

Does not affect other devs and does not break anything.

## Testing

No tests, already have hunger aggro test.


StarDylan (approved):
Makes sense to me

PR #684, opened 2025-03-09
676 bug placing potion stands in front of gate ruins npc pathing

AndrewJMart:
## Description
Upon placing a Marketstand in front of a gate, this messes with the NPC pathing behavior and completely breaks any gathering behavior by NPCs.

## Problem
This change is required as players could place Marketstands in front of gates and completely destroy any NPC gathering that would occur. 

## Context
This solution presents the simplest option to solving this problem. There is no reason to abstract this issue to other items as other placed items are walkable (meaning they can be walked through).

## Impact
This change is self isolated and was appropriately tested in a unit test.

## Testing
This functionality is tested within createMartket.test.ts with a mock test automatically.

### DIFF
--- a/packages/server/src/items/uses/marketstand/createMarket.ts
+++ b/packages/server/src/items/uses/marketstand/createMarket.ts
@@ -17,6 +17,11 @@ export class CreateMarket implements Use {
   }
 
   interact(mob: Mob, item: Item): boolean {
+    // Find if Fence Gates Are Near Mob In Radius of 1, If Yes Stop Player From Placing Fence To Prevent Pathing Errors
+    if (Item.countTypeOfItemInRadius('gate', mob.position, 1) > 0) {
+      return false;
+    }
+
     return Create.createItemFrom(item, mob, this.type);
   }
 }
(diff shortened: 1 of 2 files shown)


AndrewJMart:
Terraform Server Deploy Error is unrelated to changes made in this branch.

Meaning this change did not change anything within any Terraform files.

logancostello (approved):
simple lgtm

jmdebarro (approved):
lgtm

packages/server/src/items/uses/marketstand/createMarket.ts
```diff
@@ -17,6 +17,11 @@ export class CreateMarket implements Use {
   }
 
   interact(mob: Mob, item: Item): boolean {
+    // Find if Fence Gates Are Near Mob In Radius of 1, If Yes Stop Player From Placing Fence To Prevent Pathing Errors
+    if (Item.countTypeOfItemInRadius('gate', mob.position, 1) > 0) {
+      return false;
+    }
```
StarDylan, line 23: In the future, we should throw up an error message saying we can't do that. Since right now there is no feedback for this failing to place.

StarDylan (approved):

Issue #580, opened 2025-03-03
Bug: Cannot Drop Item After Attempting to Stash With Full Inventory


irwinj22:
Per [PR 520](https://github.com/sloalchemist/potions/pull/520), a user is not able to add more than 12 items into inventory. However, after user attempts (and fails) to add a 13th item to inventory, they are unable to drop that same item back into the world.

PR #694, opened 2025-03-10
Stash Button Shouldn't Populate with Full Inventory

irwinj22:
## Description
Previously, the user was given the option to stash an item even if they had a full inventory. Now, the stash button doesn't populate when inventory is full.

## Problem
Closes [Issue #580](https://github.com/sloalchemist/potions/issues/580).

## Context
This issue was originally raised because of a bug in which the user could not drop an item after attempting to stash it with a full inventory. However, while addressing that issue I realized (with the help of @spruhaN) that it made more sense to just not provide the option to stash when inventory is full. That is what this PR accomplishes.

## Impact
This change should not have any significant effect on other developers.

## Testing
Tested manually.

### DIFF
--- a/packages/client/src/services/serverToBroadcast.ts
+++ b/packages/client/src/services/serverToBroadcast.ts
@@ -93,7 +93,6 @@ export function setupBroadcast(
     const item = world.items[data.item_key];
     const mob = world.mobs[data.mob_key];
     item.stash(world, mob, data.position);
-    world.addStoredItem(item);
     updateInventory();
   }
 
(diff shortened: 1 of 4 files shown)


irwinj22:
Also, as a note to @StarDylan: I removed line 96 of packages/client/src/services/serverToBroadcast.ts because I realized that was a duplicate line. The item is already being added to the world.StoredItems in the stash function itself (line 112 in packages/client/src/world/item.ts)

StarDylan (approved):
Looks good, thanks for the clarification about removing that line.

PR #708, opened 2025-03-11
707 fixing decision making for non walkables

ddharton:
## Description
Generalized decision making for non-walkable interactions

## Problem
Previously, this was set to specific items, this generalizes them. Previously, if a wall and fence were within a tile of the character It would not show.

## Context
This design might be slow because it creates a distance calculation for each object and then reverts them. 

## Impact
This should not break anything else.

## Testing
Manual Testing

### DIFF
--- a/packages/client/src/world/controller.ts
+++ b/packages/client/src/world/controller.ts
@@ -346,13 +346,10 @@ export function getInteractablePhysicals(
     // to get nearbyObjects, these objects needed to have had a valid position.
     // Thus, in the case that the object doesn't have a position, an error
     // should be thrown.
-    if (!object.position)
-      throw new TypeError(
-        `Expected 'object.position' to be 'Coord', but received NULL`
-      );
+    let pos = <Coord>object.position;
     return {
       object: object,
-      distance: calculateDistance(object.position, playerPos)
+      distance: calculateDistance(pos, playerPos)
     };
   });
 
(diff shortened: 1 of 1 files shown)


packages/client/src/world/controller.ts
```diff
@@ -337,39 +337,40 @@ export function getInteractablePhysicals(
...
+  let nearbyObjects = physicals.filter((p) => !p.itemType.walkable);
 
   let nearbyBaskets = physicals.filter((p) => p.itemType.type === 'basket');
 
-  if (walls.length > 1) {
-    walls = [getClosestPhysical(walls, playerPos)];
-  }
+  let objectsWithDistance = nearbyObjects.map((object) => {
+    if (!object.position)
+      return {
```
StarDylan, line 346: I would love to see a test that covers this case, since I'm not quite sure what it's for / what it does

StarDylan (changes requested):

packages/client/src/world/controller.ts
ddharton: I manually tested it, because there's no testing framework that I am aware of

StarDylan (commented):
Same msg as on slack:
So the function you were working on: `getInteractablePhysicals` just takes in Items and a Coordinate, you can test it just like this file here:
https://github.com/sloalchemist/potions/blob/fa012540f81f9f1bd78080d2ed8edb96aa9b0347/packages/client/test/items/uses/playerInteract.test.ts#L2.

Specifically, line 346 isn't covered by those tests, so I think we should add a test case that covers that situation

StarDylan (approved):

Issue #710, opened 2025-03-11
Coordinate Out of World Bounds Checker


Twicee:
Right now there's no way to check if an item is correctly defined in the world. Without a check there's potential for items to be spawned outside the bounds of the world

PR #711, opened 2025-03-11
Coordinate checker for items

emlai1263:
## Description
pnpm run create world-name will error and exit if an item is placed outside of walkable terrain
WIP - tests

## Problem
Closes issue #710 , preventing people from adding items outside of walking terrain

## Context
The code errors (with a clear edit msg if the issue) and exits instead of just skipping the item, so it's clear that the item must be moved, instead of having it skipped over.

## Impact

 

It should be easier to move items around the world without accidentally putting them in invalid places  (this doesn't include things like houses though)

## Testing

tests in ```packages/server/test/items/spawn/spawnValidation.test.ts``` to check the error displays when needed (and doesn't otherwise)

image


StarDylan (approved):
Looks good

PR #718, opened 2025-03-11
Bug Fix: Starfish asset properly configured

spruhaN:
## Description
Starfish asset would be on top of the player upon being walked over, changed it such that the Starfish asset cannot be walked over.

## Problem
General aesthetics and slight polishing of water-world.

## Context
No alternatived were considered, due to the simplistic problem.

## Impact
This should not break anything nor requires any documentation

 

## Testing
Ran the build and attempted to walk over the starfish

### DIFF
--- a/packages/json-generator/global.json
+++ b/packages/json-generator/global.json
@@ -248,7 +248,7 @@
           "value": "blobs"
         }
       ],
-      "walkable": true,
+      "walkable": false,
       "smashable": true,
       "interactions": [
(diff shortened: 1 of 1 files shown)


StarDylan (changes requested):
Please update the test files in `packages/json-generator/test` as they are currently failing. They should match what you expect in the server and client json files respectively, since you changed the global.json those need to be update those as well.

PR #721, opened 2025-03-11
Enhancement: Centered Buttons

spruhaN:
## Description
Centers buttons across all of the frame pages

## Problem
Some final polishing, the off-centered buttons just lacked polishing

## Context
Changing the width and height instead of the spacing. Though current buttons are able to display all text appropriately so just made the page less crammed by changing the spacing.

## Impact
This change does not break anything nor require documentation.

 

## Testing
Ran against the potion build until altered spacing till it seemed appropriate.

Before:
Screenshot 2025-03-11 at 4 37 01 PM
After:
Screenshot 2025-03-11 at 4 36 01 PM


StarDylan (approved):
Looks great!

PR #729, opened 2025-03-12
Fix: Replace redundant checks with meaningful validator tests

AdamMikacich:
## Description
A prior PR implemented schema validation for the `global.json file. In order to validate that the transition functioned smoothly, the generated files were compared with an expected output (the last confirmed working file state). This was to ensure that Zod did not filter out any fields that could have potentially been forgotten when writing the schema. Unfortunately, this requires both the actual global.json file and the generated global.json files to be edited when making updates. Now that the transition has been verified to not break anything over the past week, we can go ahead and remove these tests with a more flexible approach.

## Problem
Any updates to the `global.json` file also require the same updates to be made in the `expectedClient.json` and `expectedServer.json` files. This redundancy slows down development and may confuse other developers contributing to the `global.json` file.

## Context
This new approach replaces the explicit comparison check with a more flexible validation implementation. Now, the single `global.json` input file is parsed with the zod validator. Additionally, two other checks validate that the correct number of root-level keys appear in the respective client and server global.json files generated by the package. This way other developers can modify the main global.json file while still ensuring they do not violate the expected schema of the document.

## Impact
This should ease developer contributions to the `global.json` file as they only need to modify it in one place now. Additionally, if they make any mistakes, the new tests will validate that the schema has been violated and inform the developer of what changes need to be made.

## Testing
This change validates the `global.json` file matches the expected schema. Additionally, two more tests ensure that the generated client and server `global.json` files contain the expected amount of keys to validate client_breakup and server_breakup work as intended.


StarDylan (approved):
Looks good, make sure to tell other developers when you merge that they can just merge main into their branch and not worry about updating the expected Client and server.json's. Since I expect this to cause many merge conflicts.

PR #732, opened 2025-03-12
Fix: removed redundant stand smash/destroy for potion stands

jhuanny:
## Description

Removed all instances of 'destroy stand' in JSON files

## Problem

The existing 'destroy stand' button performs no functionality and the desired functionality can be achieved with the existing 'smash potion stand button'

## Context

Could have removed 'smash potion stand' and added the functionality 'destroy stand' but requires extra work when an existing button with the desired functionality already exists. 

## Impact

 

No breaking changes made

## Testing

Went into the world and made sure breaking potion stands both with and without potions was possible. 

example of destroy stand and smash potion stand buttons
image

### DIFF
--- a/packages/client/world_assets/global.json
+++ b/packages/client/world_assets/global.json
@@ -1155,18 +1155,6 @@
               "comparison": "greater_than"
             }
           ]
-        },
-        {
-          "description": "Destroy stand",
-          "action": "destroy_stand",
-          "while_carried": false,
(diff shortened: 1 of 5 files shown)


Twicee (commented):
Looks good! However, it looks like you left the screenshot portion in the description of the the PR even though you don't have one

StarDylan (approved):

Issue #722, opened 2025-03-11
Disturbing blank screen on game entry (need loading indicator)


AlfredMadere:
# Problem 

This funky green color fills the screen for a disturbingly long time before the game actually starts once you push "start". In some cases it takes so long that it appears that the game isn't actually working. I consider this to be a bug. 

<img width="823" alt="Image" src="https://github.com/user-attachments/assets/a42fdbc8-dff3-4386-8824-ca65e46e7513" />

# Solution

Add a loading screen that displays while the client is waiting for the server.

PR #746, opened 2025-03-12
Fix: Add loading bar for world scene and optimize load time

AlfredMadere:
## Description
Loading the world scene was taking forever and it wasn't obvious to users whether it was broken or if it was actually doing something. This pr speeds up the loading time by loading compressing the large mp3 files in addition to a loading indicator to make it clear that the game is actually loading.

## Problem
 #722 

## Context
We determined that 90 percent of the time during loading was spent getting cosmic_ambient.mp3, so we compressed the file with little quality degradation. The loading indicator is a real indication of how many files have been successfully loaded. 

## Impact
Users will feel more satisfied that they can see the progress of their game loading.

## Testing
We determined that load times are now about half of what they were with manual tracing and logging.

Screenshot 2025-03-12 at 12 59 30 AM


StarDylan (approved):
Looks great!

PR #753, opened 2025-03-12
Improve Chatting System for Future Development

### LINKED ISSUE #715: Bug: mobs that should not be chatted with can be chatted with other than blobs
Right now, the only mobs that cannot be chatted with are blobs and self. This is hard coded and should be turned into a broader system where a mob should only be chattable if the project managers say they are chattable

MorrowCameron:
## Description
Improve the chatting architecture to use an attribute rather than hard coding.

## Problem
Future proofs the code to be more attribute based. This will allow a more modular set of chatting conditions.

## Context
I decided on this approach as using an attribute allows for flexibility. Additionally, using a hard coded list to determine which mobs can/can't be talked to by default allows us to update the list as new mob types are added in a single location. However, if we wanted a blob npc you could talk to, that is still changeable through their attribute.

## Impact
Nah

## Testing
- pnpm test
- Created new test


NickPerlich (approved):
You cannot see blobs in the chat tab, and the system creating that effect is now more easily expandable to other mobs

StarDylan (approved):

PR #756, opened 2025-03-12
713 bug can fight villagers

MorrowCameron:
## Description
Update fight logic so only blobs can be fought. This is done through an attribute for future proofing.

## Problem
It does not make sense to be able to fight everything, only monsters should be fought. This means a new attribute should be added to mobs to make them marked as opponents.

## Context
This approach allows for future mob types to be created as fightable with the default being a friendly npc/player. By doing this it allows for changes on a per mob basis, including pvp potential.

## Impact
No doc updates required

## Testing
pnpm test
added new unit test


NickPerlich (approved):
You can no longer see villagers in the fight tab.

StarDylan (approved):

PR #768, opened 2025-03-13
Crafting inventory integration

### LINKED ISSUE #579: Crafting Potions & Inventory Integration
Right now, the player has to hold an ingredient in order to add it to the cauldron. Additionally, the inventory only holds the most recently brewed potion and drops the others. The player should be able to interact with their inventory to choose what ingredient gets mixed in the potion, rather than holding it. 

1. Be able to click items from the inventory to drop in cauldron for potion brewing.
2. Inventory should support all the potions the player has brewed and stashed (currently only stores one potion at a time).
3. In inventory, have specific names for the potions (ie "Red Potion", "Green Potion" instead of just "Potion"). 

Image

Image

Image

Feel free to create sub-issues for each bug.

WalkinStereotype:
## Description
Items from your inventory are now buttons in the brew scene so that you can add ingredients to the cauldron directly from the inventory

## Problem
This is an enhancement so that a player doesn't have to to exit the brew scene, equip the ingredient, then reenter the brew scene before adding the desired ingredient, reducing the action from ~5 button clicks to 1.

Closes #579 

## Context
The old design forces players to leave the brew scene to equip their ingredient. There is no alternative as this was an issue posted by the project managers.

One major decision was the refactoring of interactions. 

## Impact
This change shouldn't negatively impact developers. It should be noticeable to developers, so we don't think developers will need to be alerted.

## Testing
We tested by playing the game and playing around with our inventory and brew scene.

**Note:** These screenshots do demonstrate the functionality of our changes, however, because the client would lose focus when we screenshot our browser, each screenshot is from a different instance of the game.

**Inventory:** Sunflower, **In hand:** Heartbeet
image

_**Click 'Start crafting' in actions/items tab**_
image

**Inventory:** None, **In hand:** Heartbeet
image


StarDylan (changes requested):
Please make sure you pass CI before moving your PR from draft

StarDylan (approved):
Looks great!

